PR #7, opened 2015-06-23
some preliminary defaults for the races, changed willpower to just will

ChrisHC:
Added some defaults as per the design doc.
Changed 'willpower' in characters.py to just 'will' as discussed in the irc channel.

### DIFF
--- a/plugins/ainneveraces/dwarf.py
+++ b/plugins/ainneveraces/dwarf.py
@@ -5,3 +5,10 @@ class Dwarf(Race):
     def __init__(self):
         super(Dwarf, self).__init__()
         self.name = "Dwarf"
+        self.size = "small"
+        self.limbs = []
+        self.slots = {}
+        self.foci = ['brawn', 'resilience', 'alertness']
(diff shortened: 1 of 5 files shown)


plugins/ainneveraces/dwarf.py
```diff
@@ -5,3 +5,10 @@ class Dwarf(Race):
     def __init__(self):
         super(Dwarf, self).__init__()
         self.name = "Dwarf"
+        self.size = "small"
+        self.limbs = []
+        self.slots = {}
+        self.foci = ['brawn', 'resilience', 'alertness']
+        self.feats = ['heat vision', 'poison resistance', 'dark vision', 'improved climb', 'fear resistance']
+        self.bonuses = {'willsave': 1}
```
whitenoiseoss: This willsave should be will

whitenoiseoss:
I have communicated with ChrisHC to also remove limbs and slots from all the specific races, and leave only slots in Race, defined in a typical humanoid fashion. This will become the default for everything. @ergodicbreak can speak to what the slots should be, as the OA character sheets do not specify.

Issue #3, opened 2015-06-23
Flesh out Character typeclass


whitenoiseoss:
The Character typeclass is incomplete and needs a great many more methods.

At the time of this issue being written, it is only capable of being skeleton'd out for the most part. It should return certain properties (such as armor) and have any methods that it may need. Keep in mind Character is the base for NPC/Mob/etc. Things specific to things like Shopkeepers and beasts should not go in Character.

While the programming is not difficult, the amount of other things that must be considered is fairly formidable.

Nadorrano:
Hi, I can help with an equip handler that has a number of advantages over the use of a simple dict, error checking, reporting items in the right order (from head to feet) and more. Can be easily expanded to collect modifiers from objects, stats etc. I attached it to the character typeclass as a property (evennia @lazy_property works too, but I don't know if it's risky), slots are loaded at init from a list and the equip is saved as obj.db.equip (a dict).
Bonus: equip, wear, wield (equip primary), hold (equip secondary), remove commands. The wear command is dynamic, it changes to wield or hold if the object has primary or secondary hand as slot.
My commands assume the object has the lock 'equip' and a 'slot' attribute which is a string or a list of strings. Primary and secondary hands are hardcoded as 'right hand', 'left hand' at the moment, assuming all characters right-handed.
If you need this I will fix and write a documentation for it.

whitenoiseoss:
Nadorrano, that sounds like some decent code you have there. However, you're primarily talking about equipment handling, and unfortunately we do not yet have any equipment Typeclasses coded, so that would make those pieces rather difficult.

I also think 'equip' should be part of a Player's Command Set, but the on_equip trigger I would like to put on the weapon typeclass.

I can create issues for writing out the Item typeclass, and from Item will come Weapon, Armor and Shield, if you would like to work on getting combat objects in the game, Nadorrano. Then you could work on a Character being able to equip them, and the methods for showing off a Character's equipment.

whitenoiseoss:
The issue for creating Item is now here: https://github.com/evennia/ainneve/issues/10

Nadorrano:
My at_equip hook at the moment is on the base object typeclass, however I prefer it on the Item typeclass, is more general than weapon I think.
And also why players? You don't want mobs to have an equipment?

See you at #10

ergodicbreak:
I think we have this in hand, we can open more specific issues if necessary.

PR #11, opened 2015-06-24
added some basic methods for primary and secondary traits

ChrisHC:
This my try at Issue #3.

I modeled the methods to change base and mod values of the traits as discussed with whitenoise.

Mana gets a special treatment because the max is capped at 10 as per the blue rulebook (page 49).

### DIFF
--- a/typeclasses/characters.py
+++ b/typeclasses/characters.py
@@ -75,142 +75,70 @@ class Character(DefaultCharacter):
 
         self.db.slots = {}
 
-    def base_stat(self, stat, amount):
-        if not stat in [
-                'strength',
-                'perception',
-                'intelligence',
-                'dexterity',
-                'charisma',
-                'vitality',
-                'magic',
-                'health',
-                'stamina',
-                'skills',
-                'languages',
-                'fortitude',
-                'reflex',
-                'will',
-                'mana',
-                'armor']:
-            return False
+    # helper method, checks if stat is valid
+    def find_stat(self, stat):
+        if stat in self.db.primary_traits:
+            return self.db.primary_traits[stat]
+        elif stat in self.db.secondary_traits:
+            return self.db.secondary_traits[stat]
+        else:
+            return None
 
+    # helper method, checks if race gets extra language points
+    def determine_language_points(self):
+        if self.bonuses['languages']:
+            return self.bonuses['languages']
+
+    def base_stat(self, stat, amount):
+        # sets the secondary traits
+        if stat == 'vitality':
+            base_stat('health', amount)
+            base_stat('stamina', amount)
+            base_stat('fortitude', amount)
+            base_stat('reflex', amount)
+
+        if stat == 'intelligence':
+            bonus_language_points = determine_language_points()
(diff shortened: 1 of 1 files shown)


typeclasses/characters.py
```diff
@@ -91,6 +75,71 @@ def str(self, amount):
...
+
+    # helper method, checks if race gets extra language points
+    def determine_language_points(self):
+        if self.bonuses['languages']:
+            return self.bonuses['languages']
+
+    def base_stat(self, stat, amount):
+        # sets the secondary traits
+        if stat == 'vitality':
+            base_stat('health', amount)
```
whitenoiseoss: Using base_stat within base_stat is recursive functionality. Python does not do exceptionally well with recursion, and iteration is preferred.

This should look something like this instead:

``` python
if stat == 'vitality':
   for secondary in ['health', 'stamina', 'fortitude', 'reflex']:
      secondary.base = amount
```

typeclasses/characters.py
```diff
@@ -91,6 +75,71 @@ def str(self, amount):
...
+        # as per the OA blue rulebook mana can never exceed 10 (page 49)
+        if stat = 'mana':
+            if amount > 10:
+                self.db.secondary_traits['mana'].base = 10
+                return
+            else:
+                self.db.secondary_traits['mana'].base = amount
+                return
+
+        stat_valid = find_stat(stat)
```
whitenoiseoss: `stat_valid` is a name that sounds like it would be either `True` or `False`. As it is, it is a stat object. `stat` is taken, however, so I wonder if this could be something like `valid_stat` instead.

typeclasses/characters.py
```diff
@@ -91,6 +75,71 @@ def str(self, amount):
...
+        if stat_valid:
+            stat_valid.base = amount
+        else:
+            return False
+
+    def mod_stat(self, stat, amount):
+        # as per the OA blue rulebook mana can never exceed 10 (page 49)
+        if stat = 'mana':
+            if (self.db.secondary_traits['mana'].base + amount) > 10:
+                self.db.secondary_traits['mana'].mod = 10
```
whitenoiseoss: Keep in mind `current` will return `base + mod`, so if `base + amount > 10`, then setting `mod` to 10...what if base is already 10? `current` will now be 20. This logic is not correct and will have to be thought through carefully.

typeclasses/characters.py
```diff
@@ -101,6 +150,10 @@ def become_race(self, race):
         Returns:
             bool: True if successful, False if otherwise
         """
+        # check if race is a valid races object
+        if not isinstance(race, races):
```
whitenoiseoss: This should be removed, as it will break `become_race`. `load_race` takes care of all the logic you would want here, and the `race` being passed in will be a string, so this would fail anyhow.

In addition, `races` is a module, `races.race.Race` would be the class to hit `isinstance` on.

However, this should just be removed :)

whitenoiseoss:
This is shaping up to look like good code! Thanks for your contributions, ChrisHC.

PR #13, opened 2015-06-26
Added: equip handler, commands related to the equipment, basic item typeclasses

Nadorrano:


commands/equip_commands.py
```diff
@@ -0,0 +1,258 @@
...
+
+        if not args:
+            caller.msg("Wear what?")
+            return
+
+        # this will search for a target
+        obj = caller.search(args)
+
+        if not obj:
+            return
```
ergodicbreak: Do we need a failure message here?

commands/equip_commands.py
```diff
@@ -0,0 +1,258 @@
...
+            caller.execute_cmd("wield %s" % args)
+            return
+
+        if not obj.db.slot or 'holds' in utils.make_iter(obj.db.slot)  \
+           or not obj.access(caller, 'equip'):
+            caller.msg("You can't equip %s." % obj)
+            return
+
+        if obj in caller.equip.items():
+            caller.msg("You already equip %s." % obj.name)
```
ergodicbreak: Change msg to "You're already wearing %s" for grammar.

commands/equip_commands.py
```diff
@@ -0,0 +1,258 @@
...
+            return
+
+        if not caller.equip.add(obj):
+            string = "You can't equip %s." % obj
+            if [s for s in utils.make_iter(obj.db.slot) if s in caller.equip.empty_slots()]:
+                string += " You have already something there."
```
ergodicbreak: Change string to " You already have something there." for grammar

commands/equip_commands.py
```diff
@@ -0,0 +1,258 @@
...
+
+        if not args:
+            caller.msg("Wield what?")
+            return
+
+        # this will search for a target
+        obj = caller.search(args)
+
+        if not obj:
+            return
```
ergodicbreak: failure message?

commands/equip_commands.py
```diff
@@ -0,0 +1,258 @@
...
+        # call hook
+        if hasattr(obj, "at_equip"):
+            obj.at_equip(caller)
+        caller.msg("You wear %s." % obj)
+        caller.location.msg_contents("%s wears %s." % (caller.name.capitalize(), obj),
+                                     exclude=caller)
+
+
+class CmdWield(Command):
```
ergodicbreak: There seems to be much duplication of functionality between Wear and Wield. Could we perhaps internally alias Wield to Wear? That is, put the main functionality in Wear, and when wield'ing dispatch to Wear with the proper arguments for wield.

commands/equip_commands.py
```diff
@@ -0,0 +1,258 @@
...
+
+        # call hook
+        if hasattr(obj, "at_equip"):
+            obj.at_equip(caller)
+
+        caller.msg("You wield %s." % obj)
+        caller.location.msg_contents("%s wields %s." % (caller.name.capitalize(), obj),
+                                     exclude=caller)
+
+class CmdHold(Command):
```
ergodicbreak: Similarly with Hold, it shares much with Wear and Wield. Can we refactor so Wear, Wield, and Hold use functionality from a more abstract helper class/function(s)?

I'll try to make my thoughts more clear in a gist, https://gist.github.com/ergodicbreak/1caee9eee79f2688c181

typeclasses/characters.py
```diff
@@ -9,10 +9,11 @@
 """
 # import importlib
 from evennia import DefaultCharacter
+from evennia.utils import lazy_property
 from utils.trait import Trait
```
ergodicbreak: We have moved Trait to ./world/traits/trait.py

typeclasses/items.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def at_object_creation(self):
+        """
+        Setup weapon-specific variables.
+        """
+        super(Weapons, self).at_object_creation()
+        self.locks.add("equip:all()")
+        self.tags.add(self.weapon_type, category='weapon')
+
+class Shield(Armor):
```
ergodicbreak: We should think on how we want Shield to be used in Melee for attacks

commands/equip_commands.py
Nadorrano: No, it's handled by the search method.

commands/equip_commands.py
Nadorrano: Thanks, non native english here! I think all the strings in the commands should be checked.

commands/equip_commands.py
Nadorrano: See above.

commands/equip_commands.py
Nadorrano: I know, but after thinking a lot I found code duplication to be the simpler solution, and the better readable one for newbie, future developers and people using this project as a "library". If you think at it, also the default evennia Drop and Get have a lot of code duplicated while both they are just a move_to from a location to another.
However the double database search is hurting me a lot.

Key points of the problem:
- Code duplication is quite intrinsic in the problem, if you merge Wear and Wield you have to check, every time, the argument: `if self.action == 'wear': send this message else: send another`.  You will not avoid code duplication, you just duplicate the code in the Wear command making it very messy. The first version of my commands behave like this and it was a pain.
- Command aliases can't be used: commands became not independent to each other and can't be distinguished by locks. Also does not resolve the above point.
- I think Hold is a quite different command and should stay alone, I'm thinking about future expansions of this command that don't apply to wear and wield. In MUDs I played Wear did not call Holds automatically. 
- A clean solution can be to split commands in a few functional part (check, search, equip, send messages) and play with inheritance, but it is not the evennia standard and readability suffers.

So my proposal is to keep the commands like they are and passing the object typeclass (to be wield) directly to the Wield command as an argument, not as a string to be searched, something like `cmd.obj_to_wield`. When this argument is set on the wield command the database search is not called and this object is used instead.
Let me know your opinions!

typeclasses/characters.py
Nadorrano: I know, I will update my repo and update my pull request
Related question: I put equip.py in world, should this be in utils? The handler is quite modular and not strictly related to the rpg system.

typeclasses/items.py
Nadorrano: Yep, just a skeleton at the moment, functionalities can be added later, easily.

commands/equip_commands.py
ergodicbreak: Yes I think your proposal has a lot of good points. I'll think about this today but probably we'll keep the duplication.

typeclasses/characters.py
ergodicbreak: For world I've been following what the game_template says, that it is for 'things which are not commands or typeclasses'. So I think ./world is OK for equip.py. 'World' perhaps needs a better name.

PR #17, opened 2015-07-02
Overloaded return_appearance method for autoexits and better formatting

time-:
I use this method in my project to simplify navigation. It works but it will need some modding to fit new room and item typeclass. I suggest to include some sort of "room_desc" attribute (can be named better) in character and item typeclass for custom "dropped" description.

### DIFF
--- a/typeclasses/rooms.py
+++ b/typeclasses/rooms.py
@@ -18,4 +18,45 @@ class Room(DefaultRoom):
     See examples/object.py for a list of
     properties and methods available on all Objects.
     """
-    pass
+    
+    
+    # Overloading default return_appearance method to show automatic exits
+    # destinations and divide users and things. One user per line, one thing
+    # per line. It's temporary: as object typeclass is ended, the method should
+    # return the item short_desc or some "in_room_desc" (ex. A sword is dropped
+    # here)
+    
+    def return_appearance(self, looker):
+        """
+        This formats a description. It is the hook a 'look' command
+        should call.
+
+        Args:
+            looker (Object): Object doing the looking.
+        """
+        if not looker:
+            return
+        # get and identify all objects
+        visible = (con for con in self.contents if con != looker and
+                                                    con.access(looker, "view"))
+        exits, users, things = [], [], []
+        for con in visible:
+            key = con.key
+            if con.destination:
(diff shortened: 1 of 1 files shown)


Nadorrano:
Do it! I think the technical name should be "short_desc", correct me if I'm wrong.
You should add it to the main Object typeclass, could be something like:

``` python
@property
def short_desc(self):
    return self.db.short_desc
```

And add the database variable to the main Object.

``` python
class Object(DefaultObject):
    def at_object_creation(self):
        self.db.short_desc = "%s is here." % self.name
```

Then make the default character typeclass inherit from Object

``` python
import typeclasses.Object
class Character(Object, DefaultCharacter):
```

Now you can access it, in your module, with `item.short_desc`.

Edit: some not clear instructions

time-:
Thanks a lot for suggestions, I'm working on it. Actually I'm thinking about short_desc as what you see when you check inventory or, for example, you attack someone or use an item.
You create "sword long iron" (_key_) with /short switch "An iron long sword" (_short_desc_) and it gets default appearance attribute "An iron long sword is laying here" (_self.db.appearance?_). The default appearance could be controlled by some other attributes (weight? size?) and be appended automatically, choosen from an hardcoded list or why not, an ingame-command-controlled list. Of course the builders will then change it as needed, but at least we give lazy builders semi-automatic tool to fast build.
What do you think?

Nadorrano:
> Thanks a lot for suggestions, I'm working on it. Actually I'm thinking about short_desc as what you see when you check inventory or, for example, you attack someone or use an item.

For this you can use get_displayed_name(), or just obj.name, evennia handles spaces in names so it should be enough. Ask whitenoise fot the technical name of what I call "short_desc", (description of the object when in the room).

> The default appearance could be controlled by some other attributes (weight? size?) and be appended automatically

I suggest a `@property` for this reason. The code I shown should be on the base Object, for characters should be something like:

``` python
@property
def short_desc(self):
    if self.is_sleeping:
        return "%s is sleeping." % self.name
    return self.db.short_desc
```

And adding "Heavy" to a sword it just as easy.

ergodicbreak:
In traditional DIKU muds the short description is like Evennia's display name. The 'long description' for items is what you see when you walk into the room.

ergodicbreak:
Closing as time- has resubmitted this as #22

Issue #5, opened 2015-06-23
Write AinneveArchetype skeleton


whitenoiseoss:
AinneveRaces can be used as an example for this code. Whoever does this will need to write the Archetype class (new Python class) and then figure out if anything and if so what goes into the module's **init**.py.

In addition, this module will need to be tied into Plugger.

ergodicbreak:
I'm going to have a go at this, if someone has started on it please chime in.

ergodicbreak:
Now working on this in branch https://github.com/evennia/ainneve/tree/archetypes.

ergodicbreak:
@ChrisHC I was considering merging my branch to master. Would you like to make a PR with what you have for archetypes? Perhaps we could merge the more developed one or combine the good parts of both branches.

whitenoiseoss:
VoidFox also did this. I merged VoidFox's so we would stop wrecking on this issue.

ergodicbreak:
Talked to VoidFox in IRC and this was just an accidental merge of unrelated code, going to keep this issue open for now.

ergodicbreak:
Remaining todo:
- update dual archetypes with new archetype name (e.g. 'Warrior-Scout'). Simple way, add a 'name' string, fancy way, update subclass dynamically
- Tie into plugger

archetypes branch is up to date with master, someone else can take the ball on this if they'd like a nice self-contained task.

ergodicbreak:
Added naming in 34cb6249c828e57317b6d4560baa0fc49a1a0cd6

ergodicbreak:
Plugger is not done, but merging this to get the ball rolling on character creation. We can add plugger later as needed.

PR #21, opened 2015-07-06
added empty logfiles to cheat the following error

VoidFox:
IOError: [Errno 2] No such file or directory: '---------------gits/ainneve/server/logs/http_requests.log

### DIFF
--- a/world/archetypes/arcanist.py
+++ b/world/archetypes/arcanist.py
@@ -1,0 +1,21 @@
+import random
+
+from world.archetypes.archetype import Archetype
+
+
+class Arcanist(Archetype):
+    def __init__(self):
+        super(Arcanist, self).__init__()
+        self.primary_traits.update({'strength': 4,
+                                    'perception':5,
+                                    'intelligence':7,
+                                    'dexterity': 5,
+                                    'charisma': 5,
+                                    'vitality': 5,
+                                    'magic':6}
+                                   )
+        self.bonuses = {'mana': 2}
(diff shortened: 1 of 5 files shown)


ergodicbreak:
This is OK to merge but I think it really belongs upstream in Evennia no?

ergodicbreak:
Also, maybe I'm reading the history wrong, but I think we merged more from VoidFox than we wanted here? The archetypes branch got merged. @whitenoiseoss you may want to revert this. 

edit: OK, looking at #5 I see your note. But I don't think VoidFox meant to merge that branch and then include it in his PR.

Griatch:
This is already closed; but it should be noted that these empty log files _are_ created by evennia. Log files are however excluded by the .gitignore file on the game dir- this usually makes sense but we might want to make sure evennia creates these files rather than erroring out when they are not found. 
I would suggest making an issue against evennia to this effect.

VoidFox:
Sorry for the confusion! 
Didnt know my pull request would include the merging i did after it.

Didnt want to make an issue on Evennia for this, becouse its the first time ive seen this happening.
And yes, this non-issue was created by .gitignore

"evennia --init gamename"  creates server/logs/server.log
"evennia migrate"                creates the others,  -> but only if server.log exists! <-

So in the End this Problem should never show on a clean Installation.

Griatch:
As of evennia/evennia@72049eeddd09ecee9a7b9dcfb401d2f38702c75a Evennia will now give a graceful warning and tell the user to manually create the missing log folder rather than throw a traceback.

PR #22, opened 2015-07-07
Rooms basic typeclass with flags and sectors handling

time-:


whitenoiseoss:
Most of this is not necessary, unfortunately. This is done in a very DIKU style, which is an anti-pattern for Evennia development. The details can be found in my mailing list response here: https://groups.google.com/forum/#!category-topic/evennia/ainneve/wjboQ9qXy_k

whitenoiseoss:
I am going to close this at the moment.

ergodicbreak:
Reopening as I think it makes more sense for time- to modify this PR in place, and until we know we will not merge this PR I don't see a reason to close.

time-:
removed unused attributes, added baseobject.py with a BaseObject class to inherit from. added long and short desc as property (good for dynamic changes as nadorrano proposed) and touched object/character typeclasses to inherit from baseobject also.
return_appearance room method is now fixed. it can do a lot with minor effort.
hope this is fine :)

typeclasses/characters.py
```diff
@@ -15,8 +15,10 @@
 from world import races
 from world.equip import EquipHandler
 
+from baseobject import BaseObject
 
-class Character(DefaultCharacter):
+
+class Character(BaseObject, DefaultCharacter):
```
whitenoiseoss: This multiple inheritance should not be happening.

typeclasses/objects.py
```diff
@@ -11,8 +11,9 @@
 
 """
 from evennia import DefaultObject
+from baseobject import BaseObject
 
-class Object(DefaultObject):
+class Object(BaseObject, DefaultObject):
```
whitenoiseoss: This multiple inheritance should not be happening.

typeclasses/rooms.py
```diff
@@ -6,9 +6,11 @@
 """
 
 from evennia import DefaultRoom
+from utils.flags_handler import FlagsHandler
+from utils.sectors_handler import SectorsHandler
+from baseobject import BaseObject
 
-
-class Room(DefaultRoom):
+class Room(BaseObject, DefaultRoom):
```
whitenoiseoss: This multiple inheritance should not be happening.

typeclasses/rooms.py
```diff
@@ -18,4 +20,64 @@ class Room(DefaultRoom):
...
+        else:
+            self.db.max_chars = value
+    
+    # Overloading default return_appearance method to show automatic exits
+    # destinations and divide users and things. One user per line, one thing
+    # per line. It's temporary: as object typeclass is ended, the method should
+    # return the item short_desc or some "in_room_desc" (ex. A sword is dropped
+    # here)
+    
+    def return_appearance(self, looker):
```
whitenoiseoss, line 63: This is good for now.

utils/flags_handler.py
```diff
@@ -0,0 +1,58 @@
+class FlagsHandler(object):
```
whitenoiseoss: This needs to be deleted.

utils/sectors_handler.py
```diff
@@ -0,0 +1,32 @@
+class SectorsHandler(object):
```
whitenoiseoss: This should also be deleted for the moment, even if we are going to eventually implement terrain, "sector" is not the right word for it. In this implementation, I don't know that we are doing terrain negotiation.

typeclasses/baseobject.py
```diff
@@ -0,0 +1,19 @@
+from evennia import DefaultObject
```
whitenoiseoss: These properties are technically not necessary, but they make other code prettier, so I'm fine with these properties existing. However, this should be using super so all of the below multiple inheritances are not necessary. Multiple inheritance should be very scarce.

typeclasses/Rooms.md
```diff
@@ -0,0 +1,52 @@
+# Room typeclass
+### Flags and Sectors
```
whitenoiseoss: Everything in this file describes features that are either anti-pattern or not needed.

commands/build.py
```diff
@@ -0,0 +1,129 @@
+#    @property
```
whitenoiseoss: This whole file should be removed.

commands/default_cmdsets.py
```diff
@@ -16,6 +16,7 @@
 
 from evennia import default_cmds
 from commands import equip_commands
+from commands import build
```
whitenoiseoss: If the above whole file is removed, these lines must be removed as well.

typeclasses/baseobject.py
whitenoiseoss: In addition, we don't need "typeclasses/baseobject", we already have "typeclasses/objects.py" with an Object class in it. Thing should inherit from that typeclasses.objects.Object and not DefaultObject in Evennia.

typeclasses/rooms.py
```diff
@@ -18,4 +20,64 @@ class Room(DefaultRoom):
     See examples/object.py for a list of
     properties and methods available on all Objects.
     """
-    pass
+    
+    # Sectors handled by sectors handler
+    @property
```
whitenoiseoss, line 34: This property should be removed.

typeclasses/rooms.py
```diff
@@ -18,4 +20,64 @@ class Room(DefaultRoom):
...
+    def sector(self):
+        return SectorsHandler(self)
+    
+    # Flags handled by Nadorrano's handler
+    @property
```
whitenoiseoss: This property should be removed.

typeclasses/baseobject.py
time-: Ok I did all the cuts, but I'm stuck here. I don't know how to set these properties on objects.Object, let all the typeclasses inherit from there, but without doing multiple inheritance. If I change the parent class, they won't get any of the defaults from the evennia code I guess, and super can be used only on direct parent class, right?
sorry mate I'm too noobie to solve this. For now I just doubled the code, adding properties into Room, Character and Item typeclass

typeclasses/rooms.py
```diff
@@ -7,7 +7,6 @@
 
 from evennia import DefaultRoom
```
whitenoiseoss: Room should be pulling from typeclasses.object.Object, and then you don't need the property code anymore. All 3 of these could be consolidated with a proper use of super. DefaultCharacter inherits from DefaultObject and just does some extra stuff to it. Object should have the properties, and Character and Room can both pull from Object. Character will need its locks.

typeclasses/characters.py
```diff
@@ -40,7 +40,22 @@ class Character(DefaultCharacter):
                     has connected" message echoed to the room
 
     """
-
+    @property
```
whitenoiseoss: See comment on Room for this. These properties should only be in one place.

PR #25, opened 2015-07-09
Slots correctly added to equip. Added invetory to character. Improved…

Nadorrano:
… items hierarchy. Added TwoHandedWeapons.

### DIFF
--- a/typeclasses/items.py
+++ b/typeclasses/items.py
@@ -32,6 +32,7 @@ class EquippableItem(Item):
 
     """
     slot = None
+    slot_operator = 'OR'
 
     def at_object_creation(self):
         """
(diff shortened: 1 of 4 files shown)


typeclasses/items.py
```diff
@@ -77,13 +82,35 @@ def at_object_creation(self):
...
+class Shield(Armor):
+    """
+    This class implements a basic shield object,
+    a functional type of armor.
+
+    """
+    armor_type = 'shield'
+    slot = 'wield2'
+
+class BreastPlate(Armor):
```
whitenoiseoss: Breastplate is one word, no capital P. That would mess people up most likely.

PR #29, opened 2015-07-16
Added basic character prompt

abbacode:
Change log:
- Prompt is automatically set upon character creation
- Prompt is automatically shown upon character move
- Prompt is automatically shown upon character login
- Prompt can be set via the 'prompt' command

Todo:
- Show the prompt upon entering of <return> command
- Move the prompt to ndb to improve performance?

### DIFF
--- a/commands/command.py
+++ b/commands/command.py
@@ -83,7 +83,7 @@ class Command(BaseCommand):
         """
         This hook is called after `self.func()`.
         """
-        pass
+        self.caller.show_prompt()
 
 
 class MuxCommand(default_cmds.MuxCommand):
--- a/commands/default_cmdsets.py
+++ b/commands/default_cmdsets.py
@@ -16,6 +16,7 @@ own cmdsets by inheriting from them or directly from `evennia.CmdSet`.
 
 from evennia import default_cmds
 from commands import equip_commands
+from commands import general_cmdset
 
 class CharacterCmdSet(default_cmds.CharacterCmdSet):
     """
@@ -34,6 +35,7 @@ class CharacterCmdSet(default_cmds.CharacterCmdSet):
         # any commands you add below will overload the default ones.
         #
         self.add(equip_commands.EquipCmdSet())
+        self.add(general_cmdset.GeneralCmdSet())
 
 class PlayerCmdSet(default_cmds.PlayerCmdSet):
     """
--- a/commands/equip_commands.py
+++ b/commands/equip_commands.py
@@ -2,7 +2,7 @@
 # Commands and cmdsetRelated to the equip handler
 #
 from evennia import CmdSet, utils
-from command import Command
+from evennia import Command
 
 class EquipCmdSet(CmdSet):
 
--- a/commands/equip_commands.py
+++ b/commands/equip_commands.py
@@ -2,7 +2,7 @@
 # Commands and cmdsetRelated to the equip handler
 #
 from evennia import CmdSet, utils
-from evennia import Command
+from command import Command
 
 class EquipCmdSet(CmdSet):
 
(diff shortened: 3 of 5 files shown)


ergodicbreak:
Given that `show_prompt` is basically a helper for `CmdPrompt`, I don't see why it couldn't be on `CmdPrompt`.  This way prompt functionality is more cleanly separated. Are there good reasons to keep it as is?

ergodicbreak:
I think we can use https://github.com/evennia/ainneve/blob/master/commands/default_cmdsets.py#L20 instead of creating a new `general_cmdset.py`

abbacode:
Outstanding issues at the moment are:
- Figure out how to get CmdNoInput command to work if we use the Ainneve Command Class.  It only seems to work if we use the Evennia one.  

class CmdNoInput(EvCommand):
    "Task to perform if the player hits <return> with no input"
    key = syscmdkeys.CMD_NOINPUT
    def func(self):
        self.caller.show_prompt()
- How do we make the command prompt show when a user executes a default command, like look.  At the moment that's part of the Evennia command set.  Preference is not to monkey patch.
- Do we create a new config command, and make 'config prompt <prompt_msg>'  a subset of it?
- How do we handle repetitive calls to the caller.db?  Should task like these be moved to caller.ndb?

ergodicbreak:
@abbacode could you explain a bit more about repetitive calls to the caller.db -- do you mean to construct the prompt?

ergodicbreak:
Also @abbacode did you see this https://github.com/evennia/evennia/wiki/Command%20Prompt ?

abbacode:
@ergodicbreak I've seen the wiki page and incorporated the suggestions there, i.e. what happens if the player hits <enter> with no input.

As for the repetitive call, I just meant that the prompt is going to be displayed upon execution of every command and every move.  The stats are likely to change as well, i.e. stamina/moves and from what little I understand of the code, this would prevent the ability for Evennia to do cached reads.

I don't know what would be considered best Evennia practice for a function that would require constant reading of dynamic attribute values from the database and would be executed very frequently.

ergodicbreak:
@abbacode remember we're reading from memory (not the database) and prompts are sent on the order of seconds, because they're a result of player commands. We can profile later but for now I don't feel like it will be an issue.

abbacode:
Adding feedback from Griatch:

Hi,

Commands are  considered to be 'softcode' in Evennia's design, while examples are given, they are meant to be overridden on a per-game basis. One way to override them consistently would be to import them into gamedir/commands/command.py - something like this:
(You can refer e.g. to evennia/default/character_cmdset.py for long import lists of all the commands)

from evennia.commands.default import general

def my_at_post_cmd(self)
     # stuff

CmdLook = general.CmdLook
CmdLook.at_post_cmd = my_at_post_cmd

You can of course also override specific classes with multiple inheritance for even more control:

```
      from evennia.commands.default import general
```

class MyCommandClass(object):
    def at_post_cmd(self):
        # stuff

class CmdLook(general.CmdLook, MyCommandClass):
        pass

You then need to go into gamedir/commands/default_cmdsets.py and make sure to add your own CmdLook to the CharacterCmdSet:

from commands.command import CmdLook

class CharacterCmdSet(default_cmds.CharacterCmdSet):
    # ...
    at_cmdset_creation(self):
        # ...
        self.add(CmdLook)

I recommend importing all default commands you want to use among the defaults. Overriding can be somewhat tedious to set up, but it's highly game-specific and you only need to set it up once.

ergodicbreak:
It seems like overriding is the way to go. One thing that occurred to me today is that we may very well want special prompts when we're building, admin'ing, etc. So we may have to add this functionality to _all_ commands, Evennia and Ainneve.

ergodicbreak:
Closing this PR, @abbacode feel free to reopen when/if you have time to work on it again.

PR #35, opened 2015-07-23
Abilities

abbacode:
Gents, this change incorporates the following: 
- Character has a skills and spells attribute (dictionary) to store ability level, i.e.
  self.db.skills['kick'] = 100
  self.db.spells['fireball'] = 50
- Ability class introduced under \world\abilities to be used by both skills and spells
- Skills are stored under \world\config\skillmanager.py
- Spells are stored under \world\config\spellmanager.py

When creating skills and spells, you can define the various messages (TO_CALLER, TO_ROOM, TO_TARGET).   In the message you can use:

*caller\* = shows the caller name
*target\* = shows the target name
*ability\* = shows the ability name

This can be expanded in the future to include other shortcuts.

We can then use the following code to interpret and send the message:

from world.config.skillmanager import show_skill_msg
show_skill_msg('MSG_TO_CALLER', 'kick', caller, target)
show_skill_msg('MSG_TO_TARGET', 'kick', caller, target)
show_skill_msg('MSG_TO_ROOM, 'kick', caller, target)


abbacode:
Closing this, will submit a new clean pull request with only relevant info.

PR #36, opened 2015-07-23
Added basic ability structure

abbacode:
Added basic ability code.
1. Character will have a new skill and spells attributes added (dictionary) which is used to store the current ability level.
2. Ability class which caters for both skills and spells has been added to \world\abilities.py
3. Skills are stored in a central repository under \world\config\skillmanager.py
4. Spells are stored in a central repository under \world\config\spellmanager.py

When creating skills and spells, you can define the various attributes including the messages which are displayed when a player uses them.  The example shortcuts which have been added are:

*caller\* = shows the caller name
*target\* = shows the target name
*ability\* = shows the ability name

This can be expanded in the future to include other shortcuts.  This allows unique ability attributes to be modified in a central location.  The following function can then be used to send messages to the relevant parties: 

from world.config.skillmanager import show_skill_msg

show_skill_msg('MSG_TO_CALLER', 'kick', caller, target)
show_skill_msg('MSG_TO_TARGET', 'kick', caller, target)
show_skill_msg('MSG_TO_ROOM, 'kick', caller, target)

### DIFF
--- a/commands/general_commands.py
+++ b/commands/general_commands.py
@@ -1,0 +1,74 @@
+from evennia import CmdSet, utils
+from evennia import Command
+from world.config.spellmanager import spell_manager
+from world.config.skillmanager import skill_manager
+from evennia.utils.prettytable import PrettyTable
+
+class GeneralCmdSet(CmdSet):
+
+    key = "general_cmdset"
+    priority = 2
+
+    def at_cmdset_creation(self):
+        "Populate CmdSet"
+        self.add(CmdAbilities())
+
+class CmdAbilities(Command):
+    """
+    abilities
+
+    Usage:
+      abilities
+
+    Show the player abilities
+
+    """
+    key = "abilities"
+    aliases = ["abil","abi","spell","skill"]
+    locks = "cmd:all()"
+
+    def func(self):
+        caller = self.caller
+
+        caller.msg('********************************************************')
+        caller.msg('{y           %s current abilities {n                    ' % caller)
+        caller.msg('********************************************************')
+        self.show_ability_skills()
+        self.show_ability_spells()
+
+
+    def show_ability_skills(self):
+        caller = self.caller
+
+        HEADER_OUTPUT = ['Skill','Level','Description',]
+        table = PrettyTable()
+        table._set_field_names(HEADER_OUTPUT)
+        if caller.db.skills:
+            caller.msg('\n{wYou have learnt the following skills:{n')
+            for skill in caller.db.skills:
+                table.add_row([
+                    '{G%s' % skill_manager[skill].name,
+                    '{R%s' % skill_manager[skill].level,
+                    '{Y%s' % skill_manager[skill].description,
+                ])
+
+        table_text = table.get_string()
+        caller.msg(table_text)
+
+    def show_ability_spells(self):
+        caller = self.caller
+
+        HEADER_OUTPUT = ['Spell','Level','Description',]
+        table = PrettyTable()
+        table._set_field_names(HEADER_OUTPUT)
+        if caller.db.skills:
+            caller.msg('\n{wYou have learnt the following spells:{n')
+            for spell in caller.db.spells:
+                table.add_row([
+                    '{G%s' % spell_manager[spell].name,
+                    '{R%s' % spell_manager[spell].level,
+                    '{Y%s' % spell_manager[spell].description,
+                ])
+
+        table_text = table.get_string()
+        caller.msg(table_text)
--- a/world/abilities/ability.py
+++ b/world/abilities/ability.py
@@ -1,0 +1,25 @@
+class Ability(object):
+    """
+    Basic ability object that can be used by both skills and spells
+    """
+    def __init__(self,name=""):
+        self.name = name                    # Name of the ability
+        self.description = "define me"      # How the abilities appears in the 'abilities' command
+        self.type = ""                      # Ability type [skill/spell]
+        self.level = 0                      # Current level of the ability
+        self.max_level = 99                 # Maximum level the ability can be increased to
+
+        self.use_delay = 0                  # How long does it take to execute on use
+        self.modifier = 0                   # The modifier for this skill, i.e. damage value
+        self.affects = {}                   # What affect is applied
+
+        self.msg_to_caller = ""             # Text to show to the caller on use
+        self.msg_to_target = ""             # Text to show to the target on use
+        self.msg_to_room = ""               # Text to show to the room on use
+
+        self.mana_cost = 0                  # How much mana does it take to use
+        self.move_cost = 0                  # How much move does it take to use
+        self.hp_cost = 0                    # How much hp does it take to cast
+
+    def __str__(self):
+        return "%s" % self.name
--- a/world/config/skillmanager.py
+++ b/world/config/skillmanager.py
@@ -1,0 +1,69 @@
+from world.abilities.ability import Ability
+
+#Create the skills database
+skill_manager = {}
+
+# hide skill
+skill_manager['hide'] = Ability()
+skill_manager['hide'].name = 'hide'
+skill_manager['hide'].type = 'skill'
+skill_manager['hide'].description = 'Allows a player to hide'
+skill_manager['hide'].level = 1
+skill_manager['hide'].hp_cost = 0
+skill_manager['hide'].mana_cost = 10
+skill_manager['hide'].move_cost = 0
+skill_manager['hide'].modifier = 0
+skill_manager['hide'].msg_to_caller = 'You slip into the shadows.'
+skill_manager['hide'].msg_to_room = '*caller* slips into the shadow.'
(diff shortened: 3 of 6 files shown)


ergodicbreak:
Starting to look at this now, one thing I should have highlighted in the OA rules for you is that spells and skills are maybe too different to treat precisely the same. OA abilities are what we think of as skills. OA spells are divided by tiers (unlike abilities) and have minimum requirements based on level (also unlike abilities). See the game mechanics wiki for more, https://github.com/evennia/ainneve/wiki/Game-Mechanics. 

Most of the functionality is compatible between the two, but we need to be able to have level prereqs for spells. Perhaps a general prereq for all abilities is the way to go, but most abilities (i.e. skills) won't have any in our implementation.

commands/general_commands.py
```diff
@@ -0,0 +1,74 @@
...
+            caller.msg('\n{wYou have learnt the following spells:{n')
+            for spell in caller.db.spells:
+                table.add_row([
+                    '{G%s' % spell_manager[spell].name,
+                    '{R%s' % spell_manager[spell].level,
+                    '{Y%s' % spell_manager[spell].description,
+                ])
+
+        table_text = table.get_string()
+        caller.msg(table_text)
```
ergodicbreak, line 74: With the repeated code here, maybe simpler to have a show() func with an argument like `option='skills'`?

world/abilities/ability.py
```diff
@@ -0,0 +1,25 @@
...
+        self.use_delay = 0                  # How long does it take to execute on use
+        self.modifier = 0                   # The modifier for this skill, i.e. damage value
+        self.affects = {}                   # What affect is applied
+
+        self.msg_to_caller = ""             # Text to show to the caller on use
+        self.msg_to_target = ""             # Text to show to the target on use
+        self.msg_to_room = ""               # Text to show to the room on use
+
+        self.mana_cost = 0                  # How much mana does it take to use
+        self.move_cost = 0                  # How much move does it take to use
```
ergodicbreak, line 21: For OA rules we should change move_cost to stamina_cost.

world/config/skillmanager.py
```diff
@@ -0,0 +1,69 @@
+from world.abilities.ability import Ability
+
+#Create the skills database
+skill_manager = {}
+
+# hide skill
+skill_manager['hide'] = Ability()
+skill_manager['hide'].name = 'hide'
```
ergodicbreak, line 8: I don't want to construct dicts in this manner, please use a dictionary literal.

world/config/skillmanager.py
```diff
@@ -0,0 +1,69 @@
+from world.abilities.ability import Ability
```
ergodicbreak, line 1: I would prefer a flatter directory structure here, so delete config and put all abilities in world/abilities. All spells can go in world/spells.

world/config/skillmanager.py
```diff
@@ -0,0 +1,69 @@
...
+skill_manager['kick'].move_cost = 10
+skill_manager['kick'].modifier = 0
+skill_manager['kick'].msg_to_caller = 'You kick *target*.'
+skill_manager['kick'].msg_to_room = '*caller* kicks *target*.'
+skill_manager['kick'].msg_to_target = '*caller* kicks you.'
+skill_manager['kick'].use_delay = 1
+skill_manager['kick'].affects = {}
+
+def show_skill_msg(mode=None,skill_name=None,caller=None,target=None):
+    """
```
ergodicbreak, line 39: This can be a method on Ability. Given the code it shares with `show_spell_msg`, I think we can make this more generic to handle spells as well, and then remove `show_spell_msg`

ergodicbreak:
I've finished reviewing, thanks very much for the code @abbacode . Let me know if my comments don't make sense or you have alternate suggestions.

world/abilities/ability.py
abbacode, line 21: I will update this attribute.

world/config/skillmanager.py
abbacode, line 8: I'm not to sure how to go about this.

If I create an instance of Ability and assign it as a value within a dictionary like this: 
skill_manager = {
                             'hide' : Ability(),
                             'kick' : Ability(),
                           }

How do I go about modify specific attributes using a dictionary literal format, i.e. what would the equivalent of this be:

skill_manager['hide'].level = 100

world/config/skillmanager.py
abbacode, line 1: I will re-arrange the directory structure to reflect this.

world/config/skillmanager.py
abbacode, line 39: I tried this and it prevented me from starting up the server with the following error msg:

  File "D:\Github\ainneve\typeclasses\characters.py", line 16, in <module>
    from world.spells.spellmanager import spell_manager
  File "D:\Github\ainneve\world\spells\spellmanager.py", line 1, in <module>
    from world.abilities.ability import Ability
  File "D:\Github\ainneve\world\abilities\ability.py", line 1, in <module>
    from world.spells.spellmanager import spell_manager
ImportError: cannot import name spell_manager

It feels circular, as I have to import the skill and spell manager dictionary into the ability.py file.  And then this is reversed, as I have to also import Ability into skill and spellmanager.py files.

It won't let me start the server up.

Can you please help troubleshoot this and let me know what I'm doing wrong?

Also, does this mean we would need to execute this command using the following syntax:
     Ability().show_ability_msg('skills','kick','MSG_TO_CALLER',caller,None)
     Ability().show_ability_msg('spells','fireball','MSG_TO_TARGET',caller,target)

commands/general_commands.py
abbacode, line 74: I'll wrap it up into one, just note that skills and spells are stored in different places, so there will still be a bit of overlap, i.e. skill_manager and spell_manager, unless you can think of a better way to represent the information.

world/config/skillmanager.py
ergodicbreak, line 8: You can nest dicts, and add parameters to `Ability()` (and write `Ability`'s`__init__` to match). But my preference is to remove `skill_manager` and then set it up like 

in world/abilities.py:

{name: sneak, .....}

{name: leadership, .....}

This file will be data only.

world/config/skillmanager.py
ergodicbreak, line 39: I think we can do this without importing Ability into abilities/spells if don't try to instantiate Ability in those files. The abilities/spells files will be data only. 

Looking back at show_ability, maybe an even simpler implementation would be to merge all abilities and spells (at runtime, the files could still be separate). Then we only have to find the key? Will we ever run into namespace issues between spells and abilities? I feel like it wouldn't be hard to have unique names for all abilities and spells.

Griatch:
If you run into circular import issues you can also consider loading the data-module on demand using one of Evennia's module loaders (in evennia/utils/utils.py) rather than to make an import statement and include them into the name space. 

Evennia uses such loaders in many places, such as for lock functions, spawner prototypes and cmdsets. 
.
Griatch

ergodicbreak:
hi @abbacode, just pinging to see where you'd like to go with this; do you still have time to make changes, or would your prefer I carry on from the current PR?

abbacode:
@ergodicbreak please carry on.  I'd do it, but to be honest, I'm not sure I fully understand what's required and I'd rather not hold you back.

ergodicbreak:
@abbacode on the contrary, you've been a valuable contributor and I'd rather not lose you! Is there some information I can provide on my end to make this more understandable, or do you have some feedback on how this PR process is not working as well as it could? I'm not an expert by any means and I'd appreciate any suggestions.

world/config/skillmanager.py
abbacode, line 8: @ergodicbreak this is the part where I am not clear on.  I would appreciate if you could clarify the following questions:
- Did you want abilities to be structured as nested dictionaries in literal format?  If so, should these be stored under one umbrella dictionary similar to the following:

abilities = {
    'kick' : {a:1,b:2},
    'hide' : {a:1,b:2},
}
- I don't understand the purpose of the **init** files and how we use these in reference to abilities.
- Would we still need to import the abilities dictionary as required when we need to refer to values within the dictionary?  

Thanks, my apologies I know these are simple questions, I haven't been able to completely comprehend how this will work.

ergodicbreak, line 8: @abbacode we can write abilities as dictionary literals one after the other, like

``` python
{name: sneak, ....}

{name: leadership, ....}
```

etc

I'm looking back at the conversation here regarding `__init__` but I'm not sure about your question, I think this is a reference to https://github.com/evennia/ainneve/pull/36#discussion_r35426112 ? If so you can disregard that, it's not about `__init__` files but about the `__init__` method on Ability (which we don't have to worry about)

To refer to abilities we would import the whole abilities.py module (i.e. file)

`from world.abilities import *` or `import world.abilities` etc.

ergodicbreak:
Closing this PR, @abbacode feel free to reopen when/if you're able to work on it again.

PR #37, opened 2015-07-24
Added slots and limbs to the main race class.

Nadorrano:


ergodicbreak:
Great @Nadorrano, thanks!

whitenoiseoss:
This uses tuples and it should not, as that makes the slots and limbs immutable.

Nadorrano:
Hi whitenoise!
It's the intended behaviour. Those slots are loaded by the equip handler as starting/fallback slots from the race class, just the first time the equip is requested. From those slots the `obj.db.equip` dictionary is created and after the initial equip creation the tuple is not used anymore. To cut an arm delete the arm from `obj.db.equip` (dictionary), to restore the arm reload `obj.db.slots` (tuple).
Is this ok?

I'm back from a long summer break, I will take a look to the code in these days, which is for sure to be refined. Tell me is something has to be done! c ya

PR #38, opened 2015-08-04
makes helper functions in world/rulebook.py more concise

derNarr:
Some small changes to world/rulebook.py that makes the code imho more readable and more pythonic. Also adds some small doc strings.


ergodicbreak:
Nice @derNarr, thanks.

Issue #20, opened 2015-07-05
Create Archetype Starter Gear


whitenoiseoss:
Write and create Starter Gear for each Archetype. This can be a joint effort between a builder and a coder.

On the builder side, what is needed:
Item short and long descriptions (i.e., "This breastplate is made of a fine..." and "a fine, leather breastplate")
What slots the item will take up
Any effects of the item (is it cursed? does it boost a stat? does it give health regen?)

On the coder side:
Create a Typeclass for each abstracted type of item (a Breastplate typeclass for example, from Armor, that is coded to only take up a 'torso' slot)--then all things that only take up the Torso can be Breastplates.
Add to core typeclasses if necessary for decided item effects
Write Spawner/Prototype code to generate these items

stefanludlow:
Something like this? 
https://docs.google.com/spreadsheets/d/1tfFeUwPevYVYujLfVxOrh89Lu7ypVOTBuoZH7yLyiGs/edit?usp=sharing

Nadorrano:
The equip handler is ok with hosting objects that take up more than one slot, however commands have to be reworked a little bit to do that.

Edit: now it is supposed one slot one item, we can use Feet for boots so we don't lose the possibility to have different earrings, rings etc. These items should have slot = ['left ear', 'right ear'], with the meaning "take the first empty slot among these"

whitenoiseoss:
@kaine904 that is a good start yes

Nadorrano:
@ergodicbreak @whitenoiseoss 

Would you like this approach:

``` python
slots = ('head', 'torso', 'feet', 'legs', 'blah', 'blah')
limbs = {'head': 'head'
              'left foot': 'feet',
              'right foot': 'feet'
              'etc': 'etc'}
```

It will allow to print pretty much every possible combination of limbs/slots vs items

ergodicbreak:
@Nadorrano @whitenoiseoss I'm going to split this slots/limbs discussion off into a new issue at #24.

ergodicbreak:
Per #24, we'll have slots and limbs where limbs basically are hittable slots. Eq can take up multiple slots. I think we may want a max number on slots, so for example you could wear multiple necklaces.

ergodicbreak:
@kaine904 would you like to continue this as well?

stefanludlow:
Let's touch base on general status.

feend78:
[...]

PR #46, opened 2015-11-08
Starter equipment 

feend78:
This is an initial attempt at #20 . It's not yet complete as described in the issue, but I'd appreciate some feedback for this much. 

_Changes include:_
- Updated skeleton typeclasses for equipment. 
- Created directory **world/content/prototypes/** and modules containing prototype definitions
- Updated PROTOTYPE_MODULES in **settings.py** to include the new prototype definition modules.
- Created a batch command script **world/content/starter_equipment.ev** that will spawn one copy of each equipment prototype and dump it at your feet.

Once you run `@py me.become_race('human')` to populate your character with inventory slots, the items can be equipped and removed with the current EquipHandler commands. Testing has brought to light a couple of issues with the EquipHandler. They'll either need to be addressed separately or as part of refactoring it.

It may be premature to refactor the default **world/prototypes.py** file into multiple files, but I imagine there will eventually be a fairly large number of prototype definitions, so I went ahead with it. I'm up for returning to a single prototypes module if that is preferred.

I wanted to get some feedback before pushing ahead with item effects and abilities, but I wonder if it's even a good idea, since it looks like there are other open issues that might have an impact on the implementation. If the Character typeclass refactoring tasks should be completed before item effects/abilities, this PR could be accepted as is. Until effects and abilities are done, though, the Arcanists' gear is pretty sad.


typeclasses/armors.py
```diff
@@ -2,11 +2,90 @@
...
-from typeclasses.items import Item
+from typeclasses.items import Equippable
+
+class Armor(Equippable):
+    """Typeclass for armor objects.
+
+    Attributes:
+      toughness (int): primary defensive stat
+    """
+    toughness = 0
```
ergodicbreak: Is this meant to be a class attribute?

typeclasses/armors.py
```diff
@@ -2,11 +2,90 @@
...
+
+class Headwear(Armor):
+    """Typeclass for armor prototypes worn on the head."""
+    slots = ['head']
+
+
+class Shoulderwear(Armor):
+    """Typeclass for armor prototypes worn on the shoulders."""
+    slots = ['shoulders']
+
```
ergodicbreak: I think for a first pass at armor types we can remove `Shoulderwear` as not needed, to limit scope creep.

typeclasses/armors.py
```diff
@@ -2,11 +2,90 @@
...
+
+class Breastplate(Armor):
+    """Typeclass for sleeveless armor prototypes."""
+    slots = ['torso']
+
+
+class UnderShirt():
```
ergodicbreak: Remove for same reason as previous, also conceptually it looks like a problem if a shirt only takes an arms slot.

typeclasses/armors.py
```diff
@@ -2,11 +2,90 @@
...
+    """Typeclass for shirts worn beneath Breastplates."""
+    slots = ['arms']
+
+
+class SleevedShirt(Armor):
```
ergodicbreak, line 34: I think it's desirable to wear a breastplate and shirt at the same time, so we need to resolve the conflict of both in a torso slot if the equip code doesn't handle layering.

typeclasses/items.py
```diff
@@ -6,7 +6,56 @@
 
 
 class Item(Object):
+    """Typeclass for Items.
+
+    Attributes:
+      value (int): monetary value of the item in CC
+      weight (int): weight of the item
+    """
+    value = 0
```
ergodicbreak: similar question as above, should these be instance attrs?

world/content/prototypes/combat.py
```diff
@@ -0,0 +1,126 @@
+"""
```
ergodicbreak, line 1: Nice content work!

world/content/prototypes/wearables.py
```diff
@@ -0,0 +1,216 @@
...
+     "desc": "Smooth and shiny, these gloves look as though they haven't " +
+             "seen a day of adventuring.",
+     "weight": 1,
+     "value": 30,
+     "toughness": 0}
+
+
+## Arcanist Vestments
+
+STARTER_ARC_SHROUD = \
```
ergodicbreak: I get the concept but I think shroud might be the wrong term. Perhaps cowl?

ergodicbreak:
@feend78 thanks for the good work. 

No problem with splitting prototype files here. 

Agreed other issues should be closed before getting too deep into effects/abilities.

feend78:
Shroud => Cowl works for me.
I can easily pull UnderShirt. I agree that the arms-only thing was weird. Though in the prototypes I did make a cape that's classed as Shoulderwear. It can go too, but I thought all the little noob Scouts would be so stylish in it.

From what I've seen, the current equip code doesn't handle layering, so we'll have to figure that out at some point. It's cool that we have so many equipment slots, but I do wonder whether it'll make it tough to balance gameplay eventually, given that the OA rulebooks seem to treat armor as a single unit, not multiple item mix-n-match. The only place I see anything broken out as shirt/trousers/shoes/etc. is when discussing non-armor clothing. 

Anyway, I'll make those changes and push them shortly.

typeclasses/armors.py
```diff
@@ -2,11 +2,81 @@
...
+from typeclasses.items import Equippable
+
+class Armor(Equippable):
+    """
+    Typeclass for armor objects.
+
+    Attributes:
+        toughness (int): primary defensive stat
+    """
+    toughness = 0
```
whitenoiseoss, line 14: I agree that this should not be a class attribute.

whitenoiseoss:
Apart from toughness, this is a good contribution.

feend78:
I included toughness on the Armor typeclass because it was listed for armor in the OA manual. If we remove it, how do we represent the item's defensive power?

Edit: @whitenoiseoss, perhaps you don't mean remove the attribute entirely, but instead would prefer it be organized differently? I used the pattern that I had seen @Nadorrano use in items_old.py of setting default values as class attributes, then loading them into instance Attributes in `at_object_creation`. I thought it was an efficient way of ensuring the Attribute is present, has a default value, and is easilly overridable by subclasses (if needed). For the spawner, the prototype definitions I wrote all override the default toughness, but it seemed desirable to have a default in case a builder forgets to include it.

I used this pattern on all the typeclasses in this PR, so if this approach doesn't work, how should I rework it?

ergodicbreak:
Initially I thought it shouldn't be a class attribute (i.e. instead of an instance attribute -- it needs to be somewhere), but I changed my mind -- someone might want to change the prototype (for example, increase the toughness of a breastplate) and it makes sense to have that on the class, not as an instance attribute, doesn't it?

Nadorrano:
Hi. I introduced the use of class attributes on items (only for slots, automatically loaded as `db` attributes) because they keep the code clean down the hierarchy. But this can generate confusion, duplication of variables and does not provide any real advantage. So I think we have to keep this to a minimum or forbid it completely.

To @ergodicbreak , I thought prototypes variables are loaded as db attributes, hence on instances, am I wrong? If wrong, ignore the first part of this comment.

feend78:
I've been reading more of the OA manuals, and the more I think about it, the more I wonder whether supporting the ability to wear multiple armor types as commonly done in modern RPGs is going to become a problem when it comes to implementing the combat system. 

I'm considering reworking this branch to include the full list of starter gear from the OA  manual, and only that gear. 

Since it looks like the basic manual does include clothing specific to individual "slots," we could change the `Armor` typeclass to take up all slots save for jewelry, etc. Non-armor clothing could then use the `Equippable` typeclass and specify slots in prototype definitons. I do wonder whether anyone would use those items, though, since they would probably only be useful for RP. 

Alternately, if we want to forego wearable clothing for now, we could also just do away with all slots on the character except "armor", "wield1", and "wield2".

Thoughts?

ergodicbreak:
I think this is a case where we design something more suitable for the mud environment; there are practical considerations for tabletop OA about armor types, targeting, etc. about how complex it should be, but since in the mud all the calculating is done by the computer we have freedom to develop a more detailed world model.

feend78:
I started working on effects and abilities for weapons and armor, and to avoid confusion, I think I'm going to withdraw this PR and re-submit it later.

Issue #39, opened 2015-08-28
'get barrel' in tutorial elicits a traceback.


feoh:
I was using a newly created player with no extra privs:

```
Traceback (most recent call last):
  File "/home/feoh/src/personal/evennia/evennia/commands/cmdhandler.py", line 417, in _run_command
    ret = yield cmd.func()
  File "/home/feoh/src/personal/evennia/evennia/contrib/tutorial_world/objects.py", line 1006, in
    func
    self.obj.produce_weapon(self.caller)
  File "/home/feoh/src/personal/evennia/evennia/contrib/tutorial_world/objects.py", line 1063, in
    produce_weapon
    wpn = spawn(WEAPON_PROTOTYPES[prototype], prototype_parents=WEAPON_PROTOTYPES)[0]
  File "/home/feoh/src/personal/evennia/evennia/utils/spawner.py", line 189, in spawn
    for key, val in all_from_module(prototype_module).items() if isinstance(val, dict)))
  File "/home/feoh/src/personal/evennia/evennia/utils/utils.py", line 803, in all_from_module
    mod = mod_import(module)
  File "/home/feoh/src/personal/evennia/evennia/utils/utils.py", line 767, in mod_import
    mod = __import__(module, fromlist=["None"])
ImportError: No module named prototypes

Above traceback is from an untrapped error. Please file a bug report.
```

ergodicbreak:
This bug looks like an effect of the reorganization of Ainneve's world dir, so Evennia is not finding the default prototype module path. Can you confirm this works on vanilla Evennia without using Ainneve?

feoh:
@ergodicbreak That's correct, when I run through the tutorial in Evennia proper (No Ainneve) I don't get the traceback.

thinkt4nk:
+1

feend78:
It seems to me that we probably don't want to encourage people to build the Evennia tutorial world in the first place. A solution for this would be to modify BASE_BATCHPROCESS_PATHS in settings.py to no longer include the default Evennia locations, and specifically only the **world** directory instead.

That's a super easy change, but then we would need to somehow update the description for _Limbo_ to give a customized Ainneve welcome message for admins and/or developers, and possibly a similar build instruction, once we have batch files that will build the Ainneve world. This is the part I'm unsure of. 

[...]

PR #47, opened 2015-11-09
New description for Limbo

feend78:
This is a purely cosmetic change, related to the discussion in issue #39. 

As mentioned in that issue, there is a separate pull request that will effectively fix the error that is described. However, with the default Limbo room description, the command to build the tutorial world is still front and center. I went ahead and created a new, Ainneve-specific description for Limbo that brings together some dev resources from the web and gives minimal instructions on where to find build scripts. I made the assumption that batch creation scripts would live in `world/content/`, but since there are currently no build scripts in the master, I wasn't sure whether we were intending them to be there, or just in `world/`. I'm open to whatever edits are desired. 

Note: If you want to see this description in action, you should be sure to pull the most recent Evennia core code, as there was a bug in the text2html parser that was rendering the links incorrectly. I submitted a fix and it was merged yesterday.

### DIFF
--- a/server/conf/at_initial_setup.py
+++ b/server/conf/at_initial_setup.py
@@ -26,14 +26,14 @@ def at_initial_setup():
         {YDeveloper/Builder Resources{n
           * Issue tracking: https://github.com/evennia/ainneve/issues
           * Discussion list: https://groups.google.com/forum/?fromgroups#!categories/evennia/ainneve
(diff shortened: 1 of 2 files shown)


ergodicbreak:
Thanks again, note I haven't had time to look at it in-game yet but I made a couple of small comments.

Issue #32, opened 2015-07-21
move become_race from Character to Race


ergodicbreak:
@whitenoiseoss I'd like to slim down the `Character` class to emphasize the modularity of Ainneve overall, and I think moving `become_race` to `Race` makes sense (changing the intent to something like `apply_race`). Thoughts?

feend78:
I plan to take this one on next. Now that there is a basis for character traits, it should be fairly easy to build all the necessary chargen support functionality into a world/races.py module, similar to how archetypes was done.

PR #49, opened 2015-11-27
Races

feend78:
This PR addresses Issue #32 and fleshes out a bit more of the functionality that will be needed for chargen.

_Summary of changes:_
- Created a world/races.py module, replacing the previous deeper package structure
- Removed `become_race` from the `Character` typeclass.
- Replaced it with `apply_race` module function, which applies racial and focus bonuses to the character's traits.
- Filled out Race classes with descriptive text from the OA basic manual.
- Created a `Focus` class and a dict of Focus data used by the `load_focus` factory function.
- Docs and tests

_Testing_

Testing this is a little tedious, as this is still just pieces of the chargen toolkit. To see race descriptions:

```
@py from world.races import load_race; r=load_race('human'); me.msg(r.desc)
```

To see Focus descriptions:

```
@py from world.races import load_focus; f=load_focus('agility'); me.msg(f.desc)
```

To see the effects of assigning a race and focus:

```
# reset any existing traits by calling with reset=True
@py from world.archetypes import apply_archetype; apply_archetype(me, 'warrior', reset=True)
sheet
@py from world.races import apply_race; apply_race(me, 'human', 'agility')
sheet
```


world/races.py
```diff
@@ -0,0 +1,330 @@
...
+
+
+class Human(Race):
+    """Class representing human attributes."""
+    def __init__(self):
+        super(Human, self).__init__()
+        self.name = "Human"
+        self.plural = "Humans"
+        self.size = "medium"
+        self.desc = ("Humans are the most wide-spread of all the races. The "
```
ergodicbreak: wide-spread s/b widespread

world/races.py
```diff
@@ -0,0 +1,330 @@
...
+class Dwarf(Race):
+    """Class representing dwarf attributes."""
+    def __init__(self):
+        super(Dwarf, self).__init__()
+        self.name = "Dwarf"
+        self.plural = "Dwarves"
+        self.size = "small"
+        self.desc = ("Dwarves are short, stocky demi-humans with long, "
+                     "respectable beards and heavy stout bodies. Their skin "
+                     "is earthen tone and their hair black, gray or dark "
```
ergodicbreak: tone s/b toned

ergodicbreak: Apologies that my only line comments were grammar policing! :) Great work on all the tests, thanks @feend78 .

What are your thoughts on slots/limbs with this PR? Do we wait for a refactor of eq to develop that or is the intent to decrease available slots?

feend78:
@ergodicbreak, no problem on the grammar points. I copied and pasted that text straight out of one of the OA manuals and didn't proofread it, so I take no offense at your edits.

As for the slots/limbs thing, I have another branch in progress that includes reworking parts of the equip handler. I plan to leave its ability to handle limbs and multiple slots, but at least for the short term in Ainneve, I was hoping we could start with the single-slot OA approach, and make enhancements like multi slot armor after we have the base OA combat system in place. 

My personal goal, if no one else jumps in on Ainneve development in a significant way, is to try to have a fully working first draft of the chargen process, as well as roughed-in combat and character leveling systems, maybe a skill-based action as an example, and a good set of builder-focused wiki pages by the end of the year. Basic as they may be, I'd prefer to have our first iteration stay as true as possible to OA rules, if only because it gives a well-defined spec to work toward. I also think it will allow us to produce something playable faster than if we try to include enhancements in the first round.

ergodicbreak:
Enhancing the base sounds like a good plan to me. This PR looks complete then, thank you!

Issue #31, opened 2015-07-16
Design the ability system


abbacode:
Hi all,

Creating this so we can have clarity around how the ability system will work.
## Objectives from IRC discussion
- Ability system based on Open Adventure rules
- Abilities can be either skills or spells, unique structure for each
- Players will start with all abilities during character creation
- Players will have ability points to distribute during character creation
- Ability points can be used to improve abilities in order to customize characters effectiveness

If this is incorrect, please advise and I'll update this post.
## Questions
- Discussion Point 1: Should players have the ability to learn and unlearn skills in the game
- Discussion Point 2: How to store abilities
- Discussion Point 3: Data driven vs de-centralised system
## Decision Point 1: Learning/Unlearning Skills

Should we be introducing the basic concept of learning/unlearning skills as part of the initial Ainneve role out? Is it an important core feature that people who fork Ainneve in the future will likely use, i.e. allow players to learn a new skill via a 'trainer mob' or purchase new 'spellbooks' etc. 
## Decision Point 2: How to store abilities

How do we actual store skills and spells on a character?  The logical conclusion would be to create a new attribute on the character class, i.e. 

self.db.skills = {}
self.db.spells ={}

New Skill and Spell classes can be created which inherit from a parent Ability Class.  These objects can then be assigned for each ability and stored under the character class.  The problem I can see with this system is how you maintain it moving forward.  Each object that is initialized using this approach is a unique instances assigned to each player.  There is no simple and clean method that allows us to make tweaks or modifications to skills or spells and have these propagate to all players.
## Decision Point 3: Data driven vs de-centralised

How do we structure the code?  The two options that have been discussed are data driven and decentralization. 

Data driven implying that we have a skill and spell database where we store all the the core attributes.  This would include things like descriptions, cast/use messages, modifiers, delay_on use settings etc.  When coding new skills and spells, you can make reference back to the database to pull key information.  

De-centralised:

A class is defined for each skill and spell and all the relevant attributes are hardcoded within the relevant python source file.
## Other

Please post any other items that need to be discussed for abilities and I'll update the top post to reflect the current position.

Nadorrano:
Hi, I have little time in this period and I didn't follow the IRC discussion. I apologize if this has already been discussed.

Point 2:
This is a problem that has to be solved at a global (archetype) level, not just for skills and spells.
[...]

PR #52, opened 2015-12-02
Skills

feend78:
This is the next installment in the Ainneve chargen toolkit. In my view, it addresses Issue #31, though I think there was some other, more command-related stuff that was being discussed in that issue as well.

The skills module is fairly straightforward. It follows the general pattern of archetypes and races thus far, and so it has an apply_skills module function that instantiates a db attribute on the character with data to power a skills TraitFactory. After they are initially applied, the skills are in "chargen" mode, with "plus" and "minus" counters in the skills' extra properties. These are only meant to be used during chargen and are removed by the `finalize_skills` function. 

While working on the `skills` command, I wasn't able to get the layouts I wanted for the lists with EvTable, so I got a bit sidetracked and created the AinneveList utility class. It may be a bit of scope creep, but I tried to define its use case as different enough from EvTable's and its feature list as generally useful enough to warrant it. It leverages the python string formatting minilanguage to create various layouts of columns, and accepts a list (or list of structured dicts, see the docstring), filling the items into the columns according to the properties of the list object. I also retrofitted the `traits` command to use it.

I also created a function in the `utils.utils` module that can be used to test all of the chargen pieces in-game thus far. To use it, 

```
@py from utils.utils import sample_char; sample_char(me, 'warrior', 'dwarf', 'resilience')
```

This loads your character with the listed archetype, race, and focus and enables the `sheet/skills` command to list the character sheet and all skills. It leaves your character's skills collection in "chargen" mode.


ergodicbreak:
Nice! A quick comment before I look at this in depth, I think it would be worth it to identify exactly where EvTable is going wrong (I read the log from IRC where you discussed this with @Griatch ).

Ideally we could make EvTable work and maybe with some bug hunting or feature additions that could happen. If not, we just spend a little extra time, but I think the benefit of a narrower API for Evennia developers in general would be worth it.

feend78:
Ok, @ergodicbreak, I've got a version of this branch that uses only EvTable and achieves basically the same output. I'm going to go ahead and commit it here, since we didn't really need any of the advanced functions of the list builder for the traits and skills commands.

I have a few more ideas for it, but I'm thinking about working on it independently and maybe submitting it in the Evennia contrib area as EvList, like you had suggested in chat.

ergodicbreak:
This looks great @feend78 , I'll leave it open today if @whitenoiseoss wants to have a look, to merge tonight.

Issue #54, opened 2015-12-09
Feature Request: Changes to the TraitFactory


Griatch:
@feend78, @ergodicbreak, @whitenoiseoss 

I have looked more closely into the handling of Traits/TraitFactory after a user asked me about implementation details. I have some suggestions from the perspective of core Evennia: 
1. To keep consistency with Evennia's source, I suggest that the `TraitFactory` class is renamed to `TraitHandler` since it works basically like a handler in Evennia (from the outside it looks like an object on which you access member-traits). 
2. The internal property `dbobject` (for the `_SaverDict`) is potentially confusingly named since "dbobject" has a legacy use in Evennia to mean "ObjectDB" and other base django model classes. Maybe I'm the only one confused by it, but I would recommend renaming it to `attr_dict` or even `saver_dict` for clarity. 
3. That said, I don't think it's really all that necessary to have TraitFactory/Handler accept a separately created Attribute dict as input. This could just as well be a string for the name of the Attribute you want to use or, better, child-classes of handlers hard-coding the storage-name (like how Tag/AttributeHandlers do it). The reason for this is partly that it muddles the API (why should the user need to know they must initialize a Attribute dict when it's only ever used internally?) but mainly because passing the `_SaverDict` itself around is potentially not future proof. That `_SaverDicts` retain their link to the database is an implementation _side-effect_ of Attributes; Their names start with a `_` for a reason - I'm not comfortable with them being passed around and themself used as a main unit of data; Attribute internal implementation may (not saying it will, but may) change at some point. Letting the TraitFactory itself manage its own storage is also in line with how all other handlers in Evennia works (the standard is in fact to pass in the typeclass itself and have the handler access it to set up its storage as needed).  
4. There is no need to fully comply with how handlers are used in Evennia, but you should be aware that _all_ evennia handlers use handler methods named `get/add/remove/clear/all` to access and modify internal state. In Evennia's case this is because handlers generally require extensive extra arguments (so a plain property assignment is not enough). But it's worth to keep in mind. 
5. As far as I can tell there does not seem to be a reason to not use the `evennia.utils.utils.lazy_property` decorator for assigning `traits`/`skills`properties on `Character`, similarly to how it's done for `equip`.

feend78:
[...]

PR #55, opened 2015-12-10
Converted the TraitFactory class to TraitHandler

feend78:
This PR addresses Issue #54.

Implemented methods of Evennia handler interface: `add`, `get`, `remove`, `clear`, and `all` as a property returning a list of the trait keys held by the handler.

Updated character typeclass, and archetypes and skills modules accordingly.
Updated docs and added tests for new `TraitHandler` functions.

@Griatch: I went with a string argument for the storage key on the handler's constructor, rather than hard coding it in a subclass, since the value of the key and the name of the enclosing property function are the only differences between a "traits" TraitHandler and a "skills" TraitHandler.

### DIFF
--- a/typeclasses/characters.py
+++ b/typeclasses/characters.py
@@ -13,7 +13,7 @@ from evennia.utils import lazy_property
 
 from objects import Object
 from world.equip import EquipHandler
-from world.traits import TraitFactory
+from world.traits import TraitHandler
 from world.economy import GC, SC, CC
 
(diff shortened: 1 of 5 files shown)


ergodicbreak:
Thanks @feend78 I'll give @Griatch a chance to look at this if he wants to, and merge later

Griatch:
Looks ok as far as I can tell. :)

PR #56, opened 2015-12-31
Character Creation

feend78:
This address Issue #6 

Here is a first draft of a character generation menu. I tried to do my best to have it make sense, but I would like to have a few people test it out and give feedback, especially if more can be done to beef up the help messaging.

This branch includes some significant changes to the configuration, most notably changing the MULTISESSION_MODE parameter to 2, which allows players to create more than one character, and logs them into an OOC area upon connecting where they can create new characters with the `@charcreate` command.

Players are not allowed to exit the character generation menu until it is complete, but in the event of a server restart or disconnection, the `@charcreate` command will resume the menu from the appropriate point. Also, the `@ic` command has been overridden to check whether the character has a `chargen_complete` attribute set on it, and launches the menu if it is not, which disallows puppeting of incomplete characters.

_Note: there is currently a bug in Evennia which will throw an error if you connect with a non-superuser account. I have submitted a pull request to address it, but until that is merged, only the superuser can be used for testing._


commands/chargen.py
```diff
@@ -0,0 +1,190 @@
+"""
+Chargen commands.
+
+Ainneve implements MULTISESSION_MODE = 2, in which players enter an
+OOC state immediately after creation. From that state, the @charcreate
+command is available to create a new character. This command launches
+the Ainneve character creation EvMenu.
+
+This new @charcreate command is added to the session cmdset. We also
+create a cmdset with the "Remove" mergtype to remove the original
```
ergodicbreak: mergtype -> mergetype

commands/chargen.py
```diff
@@ -0,0 +1,190 @@
...
+            self.msg("You may only create a maximum of %i characters." % charmax)
+            return
+
+        # create the character
+        from evennia.objects.models import ObjectDB
+
+        start_location = ObjectDB.objects.get_id(settings.START_LOCATION)
+        default_home = ObjectDB.objects.get_id(settings.DEFAULT_HOME)
+        typeclass = settings.BASE_CHARACTER_TYPECLASS
+        permissions = settings.PERMISSION_PLAYER_DEFAULT
```
ergodicbreak, line 134: given the `@ic` command where potentialy you could puppet any object, how does this interact with the assumption of `BASE_CHARACTER_TYPECLASS`?

commands/chargen.py
```diff
@@ -0,0 +1,190 @@
...
+            # only allow creator (and immortals) to puppet this char
+            new_character.locks.add("puppet:id(%i) or pid(%i) or perm(Immortals) or pperm(Immortals)" %
+                                    (new_character.id, player.id))
+            player.db._playable_characters.append(new_character)
+
+        else:
+            if new_character.db.chargen_complete:
+                self.msg(("{name} has already completed character "
+                          "generation. Use @ic {name} to puppet.").format(
+                            name=new_character.key
```
ergodicbreak, line 164: Should we stick with a consistent string interpolation method, either `%` or `format`?

feend78, line 164: I've consistently used `format` in the code I've written. I began with the original CmdCharCreate code, which used `%`. I'll go ahead and change them.

typeclasses/chargen.py
```diff
@@ -0,0 +1,587 @@
...
+def menunode_races(caller, raw_input):
+    """Race listing menu node."""
+    text = ""
+    if raw_input and raw_input[0] == 'F':
+        text += raw_input
+
+    text += "\n\nNext, select a race for your character."
+
+    help = fill("Your race gives your character certain bonuses and"
+                "detriments, and makes certain 'focuses' available. "
```
ergodicbreak: Is there a missing space here between "and" and "detriments"?

commands/chargen.py
feend78, line 134: Good point. I can check the typeclass in  `@ic` and just pass any non-character object through to the default `@ic`.

typeclasses/chargen.py
```diff
@@ -0,0 +1,587 @@
...
+                "in your character's starting traits.")
+    options = []
+    for arch in archetypes.VALID_ARCHETYPES:
+        a = archetypes.load_archetype(arch)
+        options.append({"desc": "|c{}|n".format(a.name),
+                        "goto": "menunode_select_archetype"})
+    return (text, help), options
+
+
+def menunode_select_archetype(caller, raw_input):
```
ergodicbreak: I think we should change the `raw_input` parameter to `raw_string`, which avoids potentially confusing newbies with Python's raw_input function and follows EvMenu's own naming convention.

typeclasses/chargen.py
feend78: Yeah, there is. I'm on the fence about using `fill` in general here. In the longer menu nodes, I tended to manually format the line breaks.

typeclasses/chargen.py
```diff
@@ -0,0 +1,587 @@
...
+
+    data = []
+    for i in xrange(3):
+        data.append([_format_trait_3col(char.traits[t])
+                     for t in archetypes.PRIMARY_TRAITS[i::3]])
+    table = EvTable(header=False, table=data)
+    remaining = archetypes.get_remaining_allocation(char.traits)
+
+    text += "Your character's traits influence combat abilities and skills.\n"
+    text += "Type 'help' to see individual skill definitions.\n\n"
```
ergodicbreak: I think "skill definitions" here should be "trait definitions"?

typeclasses/chargen.py
```diff
@@ -0,0 +1,587 @@
...
+        help = "Skill allocation is a two step process. First, you\n"
+        help += "distribute three '-1' counters across your skills,\n"
+        help += "then a number of '+1' counters equal to your Intelligence\n"
+        help += "divided by 3."
+
+        options = [{"desc": sk[s].name,
+                    "goto": "menunode_allocate_skills"}
+                   for s in skills.ALL_SKILLS]
+
+        def clear_skills(c, i):
```
ergodicbreak: `(c, i)`  -> `(caller, raw_input)`?

typeclasses/chargen.py
```diff
@@ -0,0 +1,587 @@
...
+        item = prototypes[items[int(raw_input)-1]]
+    if item:
+        text = _format_item_details(item)
+        text += "You currently have {}. Purchase |w{}|n?".format(
+                    as_price(char.db.wallet),
+                    item['key']
+                )
+        help = "Choose carefully. Purchases are final."
+
+        def purchase_item(caller):
```
ergodicbreak: maybe `purchase_item` could be moved out into the `economy` module as well and brought in with `import`

typeclasses/chargen.py
feend78: Yes. Neither is used in that callback though.

ergodicbreak: Go ahead and change it to be consistent with the others, I don't think Python has a good method of writing unused params.

typeclasses/chargen.py
feend78: The only reason I can think of that we might not want to do that would be if we want the in-game NPC shops to have actual storerooms for their inventory, similar to how it's done in one of the tutorials, or have them just spawn new items as is done here.

ergodicbreak: perhaps`purchase_item` could have a `storeroom` param that specified the behavior? The potential repeated code doesn't feel right but it'd be fine to punt on it for now.

typeclasses/chargen.py
feend78: I looked at all of the callbacks throughout the menu, and nearly all of them accept only one argument, `s`, because throughout this menu, `caller` is the session. 

I like that better for callbacks, as they rarely use the argument anyway, and could easily be modified to never use it.

ergodicbreak:
@feend78 on an organizational note do we want the menu nodes in `/typeclasses`? 

Also is there some functionality in EvMenu that timeout's on a yes/no dialogue? Sometimes if I wait on a yes/no, or perhaps go to another tab in the browser and return, the yes/no will expire and the previous node text repeats. It's hard to pin down the exact conditions to reproduce and I'm wondering if this is expected.

feend78:
I think I put them in typeclasses/ because I was roughly following the NPC
Shop tutorial, and that was where the example evmenu module was located. It
makes sense to me to have it in world/, if you think that's better.

I think that timeout thing is a known Evennia or web client issue. I think
I saw @Griatch acknowledge it on the Google group shortly before the
holiday break. I didn't worry about it in connection with our menu,
figuring it'd be solved eventually.

feend78:
I found an issue with the `@ic` command. I've fixed it locally, but I also want to write some tests to cover those cases before pushing it. I'm also going to put an exception to launching the menu for the superuser character as well, so to run character creation on it, the `@charcreate` command will have to be used.

Drevarr:
Error in menu node 'menunode_examine_and_buy'. If input contains leading or trailing spaces on the equipment menus.

feend78:
Thanks @Drevarr. I just pushed a fix for the leading/trailing spaces issue, as well as the `@ic` issue I mentioned above. Thanks for testing!

ergodicbreak:
hey @feend78 , I was reading IRC logs and it looked like there's still ongoing testing with this, just wanted to check in and confirm that was the case. Would you like to merge or wait?

feend78:
Yeah, it's almost there. I have a few other changes to make as a result of yesterday's testing with my coworkers. I think we're pretty much ready to go otherwise, and once they're in, we can merge. 

Looking over the list of issues/comments from yesterday, there is only one or two little formatting clean-up things to do with the menu text. Most notably, I put a fairly long description of the Skills allocation process in the help, but I'm considering putting it directly in the text of the skills allocation node, as the whole +1/-1 counters thing seemed to trip people up the first time through the menu.

There are also a few buggy things they found with some of the equip/items code that I will address in one or more separate PRs. 

In other news: I now have coworkers pestering me to build a combat system so they can fight each other in Ainneve instead of working. LOL

feend78:
@ergodicbreak, I think I'm satisfied with where this branch is at this point. As you may have seen in IRC logs, there may still be a few cosmetic changes that we want to make, but I think they can be handled separately later. I'll leave it here for another day for you to review again if you like, and we can merge it tomorrow.

Issue #59, opened 2016-01-01
Set up base Room and Exit typeclasses


feend78:
In IRC, we've had a few discussions about implementing some baseline typeclass features for rooms and exits. I wanted to continue that here. Some topics that have come up:
- Using `ExtendedRoom` as the parent class of our `Room` typeclass
- Rooms incorporating a linear range / size that affects room capacity and combat
- Rooms having terrain types and corresponding MV costs to enter
- Use of "slow exits"
- Ability to view an exit's destination through a room's Exits

feend78:
So @ergodicbreak, I was thinking a bit more about how to handle parties, positioning, room capacity, and the linear range. I don't really like the idea of a truly one-dimensional model, with one character per space on the range field, with characters presumably "jumping over" one another. 

I think a good balance between a truly linear system and the complexity of tracking positions on a 2D map would perhaps be to specify the range field as a length and width, then consider only the length dimension for combat purposes, and only look at the width dimension for passing and formation considerations between allies.

ergodicbreak:
@feend78 that's a good idea, and would work great for tunnels vs. open fields and the like.

feend78:
Okay, so I have started a branch and made the changes to Room discussed above. I added a mv_cost property to it, which will be used by our Exit typeclass to deduct from characters' MV trait. This brings up the need to have those movement points regenerate. I assume that we'll be using Script objects to handle this, and that each character will have its own.

It makes me wonder whether we couldn't incorporate MV points into combat instead of action points. 

Though also, as I'm starting to look toward building combat, I should say that I'm not against sticking with the OA-style turn based combat system, especially given that there's a turn-based combat tutorial in the Evennia docs that could be a starting point.

ergodicbreak:
I like the idea of incorporating MV points into combat (I'd have to check the OA rules again to see if they already are a factor).

On doing a turn-based combat, I'm not necessarily opposed but I think it would change the feel from a H&S game to more of a RP game -- which may not be a bad direction to go, especially given that we've talked about including RPsystem. What do you think about that change in direction (we probably should move this to the combat issue).

feend78:
I took a look at the slow_exit module from contrib. I think we can use it for some inspiration, but for interaction with character traits in Ainneve, I don't think slow_exit by itself will work.

feend78:
Completed by PR #61

PR #61, opened 2016-01-10
updated room typeclass to inherit from ExtendedRoom

feend78:
This PR does some basic setup of our `Room` typeclass. We want to include the functionality of `ExtendedRoom`, so I've made that the parent class for `Room` and written up a wiki page with instructions for builders on how to leverage its `@detail` command.

I also incorporated some of the elements of movement cost and the combat range field for rooms.

This PR only addresses the room part of Issue #59. That issue also mentions Exits, but I think we should perhaps discuss the movement cost dynamics a bit more. According to the OA rulebook, it costs X movement points to enter a space (room, in our case) of a given type. However, OA rules incorporate maps of different scales, as well as time periods of different scales that both affect movement.

Since we have a persistent multi-player world, I'd rather not tackle implementing variation in either scale for Ainneve. So it seems we could just go with all "Tactical Time" as defined in OA. This means that characters receive their allotment of MV points once every six seconds.

Under this scheme, I wonder about the experience of trying to run, though. If we only deduct movement points, characters would be able to move through, say, six rooms within one second, then have to wait there for 5 seconds until their MV points refill. We could implement a delay on the exits to reduce this... perhaps with the delay scaled to characters' base MV trait. I'll start on a draft, but comments on the idea are welcome.


ergodicbreak:
Apologies @feend78 I haven't had much time in the past week. I think the slight delay for room movement is a good idea, and probably hardly noticeable for most players, but enables logical implementation of running/escape skills, etc. which could reduce movement times as needed.

feend78:
Drawing somewhat from the `contrib.slow_exit` module, I've modified our `Exit` typeclass to handle checking and deducting movement costs, as well as implementing a travel delay as players traverse. Both the cost and delay values are stored separately along with terrain data in the `Room` typeclass. In the recently-merged character creation branch, I added a hook function `at_turn_start` that refreshes movement points, mana points, and clears power points. In the chargen menu, I added a call to add that hook to a six-second tickerhandler after traits are finalized.

In this setup, as a character is walking around in the world, the "turn start" actions will happen based on the 6-second ticker, but then when a character enters into the turn-based combat mode, the Script object representing the combat would remove it from the ticker and begin calling the `at_turn_start` hook itself to begin each turn. When a character exited combat, the combat Script would re-add the hook to the ticker.

As for travel delays, I went with whole-integer seconds. With this configuration, we do still have the effect of being able to walk through a number of rooms all at once, then get a "Moving so far so fast has worn you out." message until the six seconds expire and MV points are replenished. This delay is something fairly easily adjusted, so I'd be interested to hear feedback on it.

_Note: The new `Exit` assumes that all rooms will be of the base `Room` type, and therefore that they will have the `mv_cost` and `mv_delay` properties. After checking out this branch, your "Limbo" room's typeclass will need to be reset before creating any new rooms, or attempts to return to Limbo through an Exit will throw errors because it will be missing those properties._

_Alternatively, clearing your database and re-running `evennia migrate` after checking this branch out will create Limbo with the correct `Room` typeclass and attributes._

feend78:
@ergodicbreak, I think this branch is looking good. If you have time to take a look at it sometime in the next day or two, I'll leave it up and plan to merge it on Thursday evening. If you won't have time, could you let me know, and I'll go ahead with it sooner?

I also have corresponding wiki changes to push whenever the merge happens.

typeclasses/exits.py
```diff
@@ -6,30 +6,76 @@
...
+                del traversing_object.ndb.currently_moving
+
+        if move_delay > 0:
+            start_msg = self.destination.db.msg_start_move or "You start moving {exit}."
+            traversing_object.msg(start_msg.format(exit=self.key))
+            # create a delayed movement
+            deferred = utils.delay(move_delay, callback=move_callback)
+            # we store the deferred on the character, this will allow us
+            # to abort the movement. We must use an ndb here since
+            # deferreds cannot be pickled.
```
ergodicbreak, line 76: thanks for the good comment here

ergodicbreak:
Thanks @feend78 , this looks good to me.

typeclasses/exits.py
feend78, line 76: Thanks, but the credit goes to Griatch. I stole the guts of his slow_exit contrib module as the basis for this, and that was an unmodified part.

PR #62, opened 2016-01-17
Items fixes

feend78:
Fixed three small issues relating to items:
- `TwoHandedWeapon` multiple inheritance was incorrect. 
- Changed attribute existence testing to use `.attributes.has()` instead of `hasattr`
- Fixed error when inventory contains only items with no damage/toughness/range attributes.

Also beefed up equip-related tests to catch these.

### DIFF
--- a/commands/equip.py
+++ b/commands/equip.py
@@ -2,6 +2,7 @@
 Item and equipment-related command module.
 """
 from evennia import CmdSet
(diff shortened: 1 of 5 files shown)


feend78:
These are bug fixes and probably don't need a lot of discussion. Merging.

PR #63, opened 2016-02-27
added rpsystem module functionality

feend78:
This PR incorporates the contrib `rpsystem` module's functionality to objects in Ainneve.

Objects, Rooms, and Characters have all been modified to inherit from the ContribRP\* classes, which enables the extended emote, pose, and recog commands from `rpsystem`.

I have also modifed the chargen menu to have players create both an `sdesc` for their character, as well as a long description during the creation process.


feend78:
oof... forgot to run tests before I pushed this. I'll fix the Travis errors later today.

ergodicbreak:
I've checked this out and it looks good, thanks @feend78 . It does highlight that we should have an IC test room so I'll see about adding that.

Issue #67, opened 2016-04-05
Build a limited, wilderness "training" area within Riverport


feend78:
Before we get too much farther on the combat and other game systems, we could use a small grouping of rooms that can serve as a testing ground. 

This will be the first permanent parts of the Ainneve game world, and it should be comprised of 5-7 rooms near the center of Riverport. To anchor this area to the maps that @kylemecklem has created, it should include the bridge near the center of the town at coordinates (36, 18) (see [Riverport.pdf](https://github.com/evennia/ainneve/files/203779/Riverport.pdf)), as well as one room along the main east-west road, and 3-5 additional rooms north of the road that will be semi-wilderness along the river shore, and can provide some low-level enemies to train against.

Any interested builders should comment on this issue if you are interested in working on this or have questions.

j3b4:
I'd love to contribute here. I presume the ideal way to contribute built areas is through Batch code.

feend78:
Hi @j3b4, thanks for stepping up!

Batch code would certainly work, but in the wiki, we have sort of emphasized using `@batchcommand` scripts instead of `@batchcode` to try and lower the learning curve for those who might be interested in building content but not necessarily learning to code. 

We have also extended Ainneve's base `Room` typeclass with some additional properties, along with builder commands to set them. See the [Ainneve Builder Docs](https://github.com/evennia/ainneve/wiki/Builder-Documentation) wiki page for examples and best practice suggestions.

ergodicbreak:
I'm going to build the _Dancing Dragon_ inn (at coordinates 28. 17 -- see the more detailed map at https://docs.google.com/document/d/10C6YXdniCYWOX0HlYtR5RW2EBlsqscJBCJnjVr_-0WQ/pub ) as a starting location for new characters. It's on the same road as the bridge.

feend78:
I think the Grassy Meadow is working well enough for the original intent of this task. We can make separate issues for other areas like the _Dancing Dragon_.

@j3b4 and @ergodicbreak: I'm quite close to finishing combat for Ainneve. I'm hoping you would be up for checking it out and giving feedback once I get a PR together. I hope things are well with you guys!

j3b4:
Hi Dan,

Yes things are well, though I've been away from Evennia for several months.
It would be nice to start working on a few projects again.

PR #70, opened 2016-06-01
"Grassy Meadow" Area

feend78:
It sounds like a few of us have been working on some initial rooms for Ainneve. This is my contribution toward issue #67. It contains 13 rooms starting with the bridge at the center of Riverport and extends to the east and north of there. The rooms are fairly sparse, and I haven't really come up with a good narrative tying them together, so I expect they will be modified as things come together. In any case, this should be a reasonable example of build script organization and an implementation of the standards described in the Ainneve Builder Docs wiki page.

One other thing, I noticed in @ergodicbreak 's recent commit that the settings file had been modified to include 'evennia.contrib' as a batch file location. I'm not sure the reasoning behind this, but it seems like it might be in order to keep the tutorial_world build script functional. I think we had a discussion about discouraging people from using tutorial_world from within Ainneve, so I have removed that location from BASE_BATCHPROCESS_PATHS.


server/conf/at_initial_setup.py
```diff
@@ -19,21 +19,21 @@
...
-        {YGetting Started{n
-          As Player #1 you can use the {w@batchcmd{n or {w@batchcode{n commands to
+        |YGetting Started|n
+          As Player #1 you can use the |w@batchcmd|n or |w@batchcode|n commands to
           build components of Ainneve, or the entire world (once it has been created).
 
-          Build scripts are in the {wworld/build/{n directory and have {w*.ev{n or {w*.py{n extensions.
+          Build scripts are in the |wworld/content/{area name}/|n directories and have |w*.ev|n or |w*.py|n extensions.
 
         """)
```
ergodicbreak, line 39: thanks for updating the ansi characters.

server/conf/settings.py
```diff
@@ -47,5 +47,5 @@
                      'world.content.prototypes_mobs',
                      'world.content.prototypes_weapons'
                      )
-BASE_BATCHPROCESS_PATHS = ['world.content', 'evennia.contrib']
+BASE_BATCHPROCESS_PATHS = ['world.content']
```
ergodicbreak, line 50: I don't remember why I kept `evennia.contrib`, probably to keep test scripts in contrib available? But I agree it seems safer to leave it out.

world/content/riverport/central.ev
```diff
@@ -0,0 +1,353 @@
...
+#
+# start in "Limbo"
+#
+@tel Limbo
+
+######
+# create the first room
+#
+@dig Kai River Bridge;riverport#01 = riverport
+#
```
ergodicbreak, line 12: Have you thought about <zone><number> aliases insofar as how we'll keep track of them between build files? I'm imagining a scenario where we'll have to reserve blocks of numbers for separate build files, which seems less than ideal.

Perhaps we shouldn't try to have an ordered numbering system at all? Or some other namespaced alias system?

world/content/riverport/central.ev
```diff
@@ -0,0 +1,353 @@
...
+# set terrain
+#
+@terrain = EASY
+#
+# set range field
+#
+@rangefield = (4, 3)
+#
+# next, set its description
+#
```
ergodicbreak, line 35: Thanks for the comments, this will make a good example zone.

world/content/riverport/central.ev
```diff
@@ -0,0 +1,353 @@
...
+@desc
+
+ You stand in the middle of an ancient |wbridge|n spanning the |gKai River|n.
+
+ The |wriver|n runs from the north to the south, and a well-traveled |wroad|n
+ runs east and west.
+
```
ergodicbreak, line 42: We should document color choices to stay consistent. We also might want to plan for color schemes, where we don't actually put color tags in descs but some other semantic markup. This can be  important for light/dark client backgrounds.

world/content/riverport/central.ev
```diff
@@ -0,0 +1,353 @@
...
+ The steady presence of this tree fills you with a sense of serenity despite
+ a dark and shady-looking grove of trees to the northwest.
+
+#
+# add tree detail
+#
+@detail tree =
+
+ The tree's huge trunk is far too big for a normal-sized human to reach their
+ arms around. It is by far the tallest structure central Riverport.
```
ergodicbreak: typo, maybe "in central Riverport"?

ergodicbreak:
Thanks @feend78 !

feend78:
@ergodicbreak 

So on the ordered numbering / id question... I think you're right, that having some sort of global or zone-based numbering system that crosses build scripts would be cumbersome. However, I think we need to include some unique alias for rooms to enable stuff like the `@open` commands in this PR where the "Grassy Meadow" is made up of 7-8 rooms all named "Grassy Meadow". While we could go with specifying "1-Grassy Meadow" or similar, it is nice to have a hard coded alias that can easily be found in the batchcommand script file.

Perhaps we don't worry about having unique numbered ids on the zone or region level, but allow each batchcommand script to define its own sequence. So for example, instead of riverport#01-riverport#13 in this PR, since this file is "central.ev", they would be rp_central#01-rp_central#13 or similar (adding the 'rp' since just 'central' could conflict with another area's "central" build script).

If this approach works for you, I'll go ahead and modify my ID aliases to rp_central#00 format

world/content/riverport/central.ev
feend78, line 42: I agree with staying consistent, and looking at this line above, there may be some instances of `|r` and some of `|g`, so that should be addressed.

On the [Builder Docs](https://github.com/evennia/ainneve/wiki/Builder-Documentation#colors) wiki page, I did start documenting the color choices that have made it into master so far. 

I do like the idea of a more markup-based color-theming system, but that seems like a more systemic change than this builder task calls for, and should perhaps be addressed as a separate issue/feature request.

ergodicbreak:
@feend78 yes an internal numbering for an area file seems like it would work, I'll do that for my own building now.

world/content/riverport/central.ev
stefanludlow: Typo should be fixed.

stefanludlow:
Would love to see the items on this that are outstanding resolved and the PR brought in.

feend78:
The tests that are failing don't appear to be failing because of any of the content of this PR, but rather from changes to upstream Evennia. In most cases, I think the tests can be updated to accommodate this.

stefanludlow:
@feend78 Yeah, saw the break. Would agree. Next steps?

feend78:
I modified the tests to jive with the updated evennia/master. Next steps are we leave the PR open for maybe a day to give @ergodicbreak a chance to review it and comment if he wants, and I'll merge it in tomorrow evening, central time, if not before.

Issue #69, opened 2016-05-30
Unable to Equip Objects


helix-0311:
https://github.com/evennia/ainneve/blob/master/commands/equip.py#L92

After initial character generation (with a purchase of equipment):

> equip lance
> There is no 'lance' here.
> You have nothing in your equipment.
> 
> inv
> You are carrying:
>  a lance polearm  A sturdy wooden rod fitted with a menacing spike,   (Damage:  4)  
>                   this lance polearm threatens even enemies at a  
>                   distance.  

---

When looking at the code, this seems to be the issue:

equip.py line 92:

```
def func(self):
    caller = self.caller
    args = self.args.strip()
    swap = any(s.startswith('s') for s in self.switches)

    if hasattr(self, "item"):
        obj = self.item
        del self.item
    else:
        **obj = self.caller.search(args) if args else None**
```

---

This else statement always seems to fire, but is not able to find anything in the character's inventory. Tested this (without full debug) by doing:

print args
print obj (after assignment, always None)
print caller
print caller.contents

This verifies that the args are as expected, the obj object is None, the caller is as expected, and the caller.contents contains the object that args should reference. Additionally, tested via the find command with the args string and was able to locate the object with no issue.

helix-0311:
Some additional testing of this item:

```
def func(self):
    caller = self.caller
    args = self.args.strip()
    swap = any(s.startswith('s') for s in self.switches)

    obj = caller.search(args, location=caller)
    print type(caller.contents)
    stuff = caller.contents[0].aliases.all()
    if stuff.__contains__(args):
        print True
    else:
        print False
```

This evaluates to True in the console, making it appear that the caller.search() method is just not inspecting the aliases.

helix-0311:
https://github.com/evennia/evennia/blob/477391b5df650d06b358f1e8587813319a77d120/evennia/contrib/rpsystem.py#L1058

This appears to be the source of the issue.

Superusers are able to access the original search method, while non-superusers are unable to. Not sure the intended purpose of this block, as it would appear to block searches unless certain permission levels are met.

feend78:
Based on Griatch's comments in the parallel issue raised in the evennia repository, the true problem is not this permissions check in the rpsystem module, but that the objects in Ainneve's starter inventory lack the `sdesc` property, which replaces the use of `key` to refer to in-game items. 

[...]

PR #71, opened 2016-06-07
Equip bugfix

feend78:
Fixed an issue where the 'equip', 'wear', and 'wield' commands could target items in a the room but not in the caller's inventory that came up while investigating issue #69


feend78:
One point of discussion: in this PR, I created a function that handles searching only the caller's contents for an item as a private module level function, but I could also see it being useful in other contexts, so perhaps its name should not begin with an underscore. Comments appreciated

ergodicbreak:
It does seem like a useful utility function. Given the `candidates` argument I'm wondering if there's some way to incorporate this without expanding the existing API, but I don't see how you'd do that with a custom error message. 

(also sorry for the long delay on replying to this!)

feend78:
I agree that it's unfortunate to create an additional `search_inventory` method, but when the character's inventory is empty, calling the default `search` with candidates=self.contents evaluates `self.contents` as False (empty list), and `search` then falls back to the default candidates list, which includes the room contents.

In my last commit, I moved the `search_inventory` function onto the Character typeclass, which I think makes more sense than as a module function in a command-based module.

stefanludlow:
Agreed.

Griatch:
@feend78 

I don't agree. I think it sounds like a bug in the Evennia search mechanism if passing it an empty list for candidates doesn't return the empty list (since there are no candidates to search). It should only default if `candidates = None` rather than `candidates = []`. I made an issue against Evennia [here](https://github.com/evennia/evennia/issues/1016).

feend78:
Thanks @Griatch. I'll keep an eye on that issue and rework this branch to just use `search`.

PR #77, opened 2016-08-11
NPC Stub

feend78:
- Stubs out a basic NPC typeclass, which gets loaded with minimal default traits and skills.
- Adds support for including `traits` and `skills` keys within spawner prototype dictionaries.
- Adds builder commands `@traits` and `@skills` to allow builders to edit corresponding attributes on NPCs.


stefanludlow:
This looks good, probably ready to merge this into master?

PR #80, opened 2016-08-19
Added the "appraise" skill

andrei-pmbcn:


commands/default_cmdsets.py
```diff
@@ -38,6 +39,7 @@ def at_cmdset_creation(self):
         self.add(chartraits.CharTraitCmdSet())
         self.add(room_exit.AinneveRoomExitsCmdSet())
         self.add(building.AinneveBuildingCmdSet())
+	self.add(skills.SkillCmdSet())
```
feend78: This looks like it could be spacing issue that would cause errors. Please double-check it.

commands/skills/skills.py
```diff
@@ -0,0 +1,9 @@
+from evennia.commands.cmdset import CmdSet
+from commands.skills.appraise import CmdAppraise
+
+class SkillCmdSet(CmdSet):
```
feend78: I'm thinking since this is likely to be the only thing in this file, we may want to consider moving the `SkillCmdSet` definition into commands/skills/**init**.py. This would simplify imports in default_cmdsets.py

commands/skills/appraise.py
```diff
@@ -0,0 +1,142 @@
+from commands.command import MuxCommand
```
feend78: To keep imports as consistent as possible, could this instead be 

`from evennia import default_cmds`

and then in the command below, 

`class CmdAppraise(default_cmds.MuxCommand):`

commands/skills/appraise.py
```diff
@@ -0,0 +1,142 @@
...
+        # prohibit further appraisal.
+
+        #caller.msg("initial lose list: " + str(caller.ndb.appr_lose))
+        #caller.msg("initial win list: " + str(caller.ndb.appr_win))
+
+        lose = caller.ndb.appr_lose
+        reroll_time = 3600 # how much time one has to wait before rerolling
+                           # the appraise skill on the same type of item
+
+        if lose.has_key(obj.name):
```
feend78: Could the win/lose dicts be keyed on `obj.id` instead of `obj.name`? I could see us expanding the functionality of this command to be able to appraise objects other than items in inventory, and we can't necessarily count on every item having a `name` property. Using `id` will give more flexibility going forward.

commands/skills/appraise.py
```diff
@@ -0,0 +1,142 @@
...
+                gc = ""
+                sc = " and "
+            else:
+                gc = " and %d GC" % coin['GC']
+                sc = ", "
+            if coin['SC'] == 0:
+                sc = ""
+            else:
+                sc += "%d SC" % coin['SC']
+            self.caller.msg("The value of %s is %s%s%s" %
```
feend78: For consistency, would you be willing to use the world.economy.format_coin function here?
Also, we've been trying to move to using {} style text format syntax and away from %s. Could that happen here too?

feend78:
Overall, this looks quite good. Looking forward, I think this could be expanded beyond just value appraisals to include weapon or armor combat stats, or even to use during combat to appraise an enemy and get some idea of HP or effects or maybe even general difficulty. These things don't need to happen for this PR unless you want to add them, but maybe just something to keep in mind.

I was wondering if you could also clean out any debugging statements that are still in there but commented out.

Another request, which may be a bit more challenging, is could you write unit tests to cover this command? I'm happy to help you set them up. As I said, they can be challenging, but automated testing is a **super** valuable thing to know how to do, and it will help us maintain high quality code in the Ainneve code base.

Thanks again for this, @andrei-pmbcn . I hope the changes I proposed are cool with you. I'm happy to be argued with about them if you disagree. :)

feend78:
So the Travis CI errors in this branch are happening because of some changes to Evennia master. I've pushed an update to the Ainneve master branch to correct them. Please merge the changes from master into your branch and commit them to update this PR and hopefully get a nice happy green checkmark.

feend78:
@andrei-pmbcn, could we make the alias for appraise be 'app' instead? In my combat branch, I had already created an alias of 'appr' for "approach". I could modify mine to be 'app' but i think that appraise will be the more-useful command and therfore deserves the shorter alias

feend78:
This is looking good. I'll leave it up until this evening, in case @ergodicbreak or @stefanludlow have any input they'd like to add, then merge it.

stefanludlow:
On my end, all looks good!

Issue #82, opened 2016-08-22
Sdescs will not evaluate between a and an based on potential {}, e.g. a yellow rat, an orange rat


stefanludlow:
When dynamic descriptions are created, there is currently nothing in place to automatically evaluate them for grammatical correctness. 

The logic should be, for sdescs: 
If the first letter of the second word in a string beginning with "a" or "an" is a consonant, then the first word of the string should be "a" 
If the first letter of the second word in a string beginning with "a" or "an" is a vowel, then the first word of the string should be "an"

Griatch suggests somewhere around region https://github.com/evennia/evennia/blob/master/evennia/contrib/rpsystem.py#L443 for a place to potentially install something or other to fix.

Griatch:
This should probably be made as a feature request against Evennia rather than against Ainneve.

feend78:
I wrote this up as Issue/Feature request [#1028](https://git.io/viJiD) in Evennia.

PR #84, opened 2016-09-06
Dynamic sdescs fix for prototypes

feend78:
This is a solution to closed Issue #82. It creates a helper function that prepends "a" or "an" as needed to all words in a list. The function has been inserted into the creation of `*_adj_1` variables, and the lambdas updated accordingly.

However, one concern I have about this approach is that the search mechanism in the `rpsystem` module's base typeclasses does not accept partial words in the sdesc, which means that, given "an ochre-furred, golden-eyed wolf", you will not be able to say 'attack ochre' and have it resolve correctly. It would instead require input like 'attack ochre-furred' to work.

Of course 'attack wolf' would always work, but in a room full of wolves, it would fall back to needing '1-wolf'-style differentiation, and it seems like the purpose of creating all these lengthy, adjective-laden sdescs would be to allow players to avoid using the #-wolf syntax for the most part. 

So I guess that's something to discuss. I saw someone in the IRC channel recently expressing their desire to have `startswith()` style matching for all parsed arguments, which is essentially what is needed here. That, or a more sophisticated prototype-generating system that gracefully handles setting the dynamic sdesc along with matching shortened-form aliases.

### DIFF
--- a/commands/tests.py
+++ b/commands/tests.py
@@ -28,7 +28,7 @@ class ItemEncumbranceTestCase(EvenniaTest):
         self.obj1.db.weight = 1.0
         self.obj2.swap_typeclass('typeclasses.armors.Armor',
                                  clean_attributes=True,
-                                 run_start_hooks=True)
+                                 run_start_hooks="all")
         self.obj2.db.toughness = 1
         self.obj2.db.weight = 18.0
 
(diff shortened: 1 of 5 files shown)


stefanludlow:
Looks like we are ready to bring this in?

PR #86, opened 2016-10-23
Combat

### LINKED ISSUE #51: Design the combat system
I wrote a preliminary document based on OA here, https://docs.google.com/document/d/1RQGGoua7Xp0-AAwb75A3MMIaDnnXLCVM3RYVCztiXiI/edit?usp=sharing

feend78:
This is a first iteration of combat for Ainneve.

There are several points at which this differs from the original [Combat Proposal](https://github.com/evennia/ainneve/wiki/Combat-proposal) document. 

This branch differs in the following ways from the proposal document:
- Under "Initiative, Surprise, and Perception", **Surprise** has not been implemented.
- Range, instead of being a linear range field, is represented as a relationship between individual combatants and can be in one of three states: "melee", "reach", or "ranged". All combatants start at "ranged" distance from each other.
- Formations are not supported, as grouping has not been implemented and is not slated to be in the v0.1  version.
- Not all combat actions were implemented. See `commands/combat.py` and `world/rulebook.py` for action details.
- Power Points have not been implemented, but they are calculated and a hook is stubbed out.
- Targeting has not been implemented.
- Status effects nor weapon effects have been implemented. You can only beat one another to death.

This branch includes a basic implementation of Death. It is essentially just a brief time out for the dead character. Dead PCs get a few messages and respawn at the "Grass Shrine".

@Griatch and I had a discussion about stance in combat messaging, and I did take his suggestion and modify the combat _action messaging_ to incorporate only director stance. I left second person messages intact in the player feedback from combat _commands_ though, because the question of consistency in using director stance is something that needs more discussion. Honestly, it's something that I hadn't really considered at the start ,so I would imagine that much of the previous work in Ainneve might need to be adjusted to have truly consistent director stance throughout.

_NOTE: Anyone with a previous instance of Ainneve who wants to test this branch should start with a fresh database and an updated copy of the evennia master branch._

Resolves #51


feend78:
I have added a wiki page describing the combat implementation in greater detail. 

It can be found here: [Combat v0.1](https://github.com/evennia/ainneve/wiki/Combat-v0.1).

Griatch:
I finally got around to test this more. Here are some things I noticed: 

- Plate mail is 5 SC while Chain mail is 15 SC for less protection. Unless it has some testing purposes the price seems wrong. 
- When looking at Ol' Scurvy, the inventory of Ol' Scurvy ("his jaws") pops up, making it look like an unfinished sentence. "His jaws" should probably just be called "jaws" and the list could be prefixed with something generic like "They are using:" to make it clear: 

       Ol' Scurvy is here.
       Wretched and fearsome-looking, the wolf growls as the fur on its back bristles at the sight of hapless travelers.
       his jaws  

- In combat, as a newbie I would intuitively expect at `range` command to be available to show me the current range `help combat` is telling me about, and maybe even allow `range +/-` as an alternative to separate approach/retreat commands.
- Since the `help combat` tells me about these actions I can do, I intuitively try the "actions" command to see which those actions are. I know (because I know how the Command system works) that I can use `help` to view the available actions, but as a newbie I wouldn't. An `actions` command that returns a list of combat-relevant commands (maybe even with the first lines from the help entry next to it) would be very useful from a usability standpoint. 
- in combat I wielded a 2-hand battle axe and inserted two attack commands. The first first action cannot be an attack though; it gives the message `Griatch does not have an appropriate weapon to attack.` Attacking on subsequent actions/turns work as normal. This does not seem to be some sort of "readying the axe action" since attacking at any time later works. 
- For presentation's sake it would be a good idea to give the wolf an alias also for "Ol" and "Ol'". 

Overall, I'm impressed at how well-polished this looks already. Apart from the minor things above I didn't see any major weirdness or strange outputs anywhere. I think going for just this small subset of Races/Arcetypes/Skills was the right choice, it feels compact yet solid and showing off the features. The EvMenu design with its colors was also clean and easy to follow.

feend78:
Thanks for the feedback, @Griatch ! In my recent commits, I have 

* Addressed the chain mail/plate mail pricing issue. Those were the actual prices from the OA rulebook, but it must be a typo. I switched them for now, and they can be adjusted later if need.

* The handling of NPC attacks, weapons, and AI is really just a placeholder that I threw together to be able to demo the combat experience. I plan to revisit AI and NPC enemies after basic combat is merged. For the present, I just renamed the item from "his jaws" to "a vicious bite" in hopes that it looks less strange than the former.

* I have added a paragraph about using the `look` command in the Combat help file. Also, when combat begins, the combat handler automatically calls `look` at the start of the first turn.

* I have added an `actions` command that gives a one-line summary of combat actions.

* I have modified the messaging when your weapon is not appropriately ranged to more specifically indicate range as the problem.

* As we discussed in IRC, use of rpsystem prevents the use of aliases, so a fix isn't that easy. I think we agreed it would be a feature/issue to be addressed in that module.

PR #87, opened 2016-10-25
Alteration of Equip Commands.

ghost:
Whilst using this code for my own project I found the Equip Commands seemed to not make use of the cmd attribute of commands which indicates the alias used. Instead it had a master equip command and separate commands for aliases which simply started the equip command with additional arguments to indicate which command was used. This code achieves the same thing but in a single command through the use of a reference dictionary. 

Additionally, when an argument isn't passed or the command fails due to the object, a list of available items in your inventory is displayed for the command. I.e. a failed "wield" command would display a list of available wieldable items in your inventory as opposed to your functionality of displaying your equipped items. My usage involved showing a list of equipped items in another command, but if you would prefer the original functionality of displaying that list when using a equip command without arguments let me know and I'll add that.

### DIFF
--- a/commands/equip.py
+++ b/commands/equip.py
@@ -23,19 +23,17 @@ class EquipCmdSet(CmdSet):
     def at_cmdset_creation(self):
         self.add(CmdInventory())
         self.add(CmdEquip())
-        self.add(CmdWear())
-        self.add(CmdWield())
         self.add(CmdRemove())
 
 
 class CmdInventory(MuxCommand):
     """
     view inventory
-
+    
     Usage:
       inventory
       inv
-
+      
     Shows your inventory.
     """
     key = "inventory"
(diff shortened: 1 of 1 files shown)


feend78:
@CloudKeeper1 

This all looks good. I would like to maintain the original functionality, though, but I like the approach of showing wearable/wieldable items too. Perhaps "wear" and "wield" with no arguments can show their options while "equip" by itself shows equipped items? 

Aside from that, could you also take a look at the failing unit tests? With your changes, the tests will probably need to be updated. The full test suite for Ainneve is getting to be a bit large, so I recommend running only the module or specific test case you're working on using `evennia test commands.tests` or similar.

feend78:
I have fixed the unit test problems in ainneve:master that were caused by fixes to upstream evennia.

Please merge the master into your branch and a number of unit test failures should be fixed, though with your changes to equip, there might need to be further adjustments.

aliceafterall:
@CloudKeeper1 Would there still be interest in getting this merged into Ainneve?

PR #90, opened 2017-01-03
Add automatic key lookup to ALL_SKILLS

jrsteensen:
Removed list from ALL_SKILLS, and replaced with .keys() to automatically iterate for keys.

This prevents the need from manually editing keys in ALL_SKILLS when new skill is added or removed from _SKILL_DATA.


jrsteensen:
Never mind!

Issue #57, opened 2015-12-31
Enhance return_appearance on Character Typeclass


feend78:
The character typeclass is currently using the  default `return_appearance` hook implementation from DefaultCharacter/DefaultObject, which isn't aware of the difference between inventory items and equipped items, and displays all of the character's items to anyone looking at it.

Ideas for improvements to `return_appearance`:
- Perception checks to give more or less detailed messages to the looker, including archetype, race, possibly indicators for health and/or stamina.
- Hide contents of character inventory from lookers.
- Show equipped items to lookers.

aliceafterall:
I've mostly completed this, but I wanted to ask, how difficult should the perception check be for each of race, archetype, health, and stamina?

For health and stamina, should it be perception, or the Medicine skill?

I've already completed the rest of it - the return_appearance hook doesn't show inventory items, only equipped items, and where they're equipped ( right arm, left arm, head, etc )

PR #96, opened 2017-07-17
Return_appearance changes

aliceafterall:
This change addresses #57 by adding a customized return_appearance to the Character typeclass.

This hook shows the name, and if the looker passes perception checks, the race and archetype like so:
**Chainsol** the **Elf** **Arcanist**

Another perception check tells the looker how healthy the target is - right now, there are four messages at 80%, 50%, anything over 0, and 0 HP.

Another check tells you how much stamina they have left, also with three messages at 80%, 50%, and anything below 50%.

Finally, the return_appearance hook also doesn't show the looker the contents of the target's inventory, and instead only shows what they have equipped. To this end, the hook uses the target's EquipHandler and the limbs the target has.

To make the limbs more readable, the limbs of the default characters have been changed to be human-readable instead of shortened: right arm instead of r_arm.

The only thing I remain unsure of is the difficulty of the perception checks, and if we should store and remember the result for a period of time, so the checks are only re-performed after a period of time, say, 5 minutes?

### DIFF
--- a/typeclasses/characters.py
+++ b/typeclasses/characters.py
@@ -141,6 +141,9 @@ class Character(ContribRPCharacter):
                 string += "They seem to be in pretty bad shape.\n"
             else:
                 string += "They're dead.\n"
+        # if we don't know their health, then just show a default message
+        else:
+            string += "They seem to be in good health.\n"
 
         if knows_stamina and self.traits.HP.actual > 0: # we check HP.actual,
                                                         # in case they're dead
--- a/typeclasses/characters.py
+++ b/typeclasses/characters.py
@@ -103,20 +103,20 @@ class Character(ContribRPCharacter):
         # These each do a skill check, but they always pass
         # if you're looking at yourself
         knows_race = skill_check(per) or looker == self
-        knows_archetype = skill_check(per, 6) or looker == self
-        knows_health = skill_check(per, 7) or looker == self
-        knows_stamina = skill_check(per, 7) or looker == self
+        knows_archetype = skill_check(per) or looker == self
+        knows_health = skill_check(per) or looker == self
+        knows_stamina = skill_check(per) or looker == self
 
         # this is the base name format - it just colors the name cyan
         string = "|c%s|n" % name
 
         # if we're adding race or archetype, add "the" after the name
-        if knows_race or knows_archetype:
+        if (knows_race and self.db.archetype) or (knows_archetype and self.db.archetype):
             string += " the"
 
-        if knows_race:
+        if knows_race and self.db.race:
             string += " {}".format(self.db.race)
-        if knows_archetype:
+        if knows_archetype and self.db.archetype:
             string += " {}".format(self.db.archetype)
 
         # There may be a more efficient way to do this,
--- a/typeclasses/characters.py
+++ b/typeclasses/characters.py
@@ -120,7 +120,7 @@ class Character(ContribRPCharacter):
         string = "|c%s|n" % name
 
         # if we're adding race or archetype, add "the" after the name
-        if (knows_race) or (knows_archetype):
+        if knows_race or knows_archetype:
             string += " the"
 
         if knows_race:
--- a/typeclasses/characters.py
+++ b/typeclasses/characters.py
@@ -15,6 +15,8 @@ from world.skills import apply_skills
 from world.archetypes import Archetype
 from world.death import CharDeathHandler, NPCDeathHandler
 
+from world.rulebook import skill_check
+
 
 class Character(ContribRPCharacter):
     """Base character typeclass for Ainneve.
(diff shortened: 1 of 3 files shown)


j-f1:
Looks like the build errors occur without your changes.

aliceafterall:
Yeah, it looks like the travis CI is down. All checks are failing on everything. :/

typeclasses/characters.py
```diff
@@ -84,6 +86,101 @@ def at_pre_unpuppet(self):
...
+        # get description, build string
+        name = self.get_display_name(looker)
+        per  = self.traits.PER
+
+        # These each do a skill check, but they always pass
+        # if you're looking at yourself
+        knows_race = skill_check(per) or looker == self
+        knows_archetype = skill_check(per) or looker == self
+        knows_health = skill_check(per) or looker == self
+        knows_stamina = skill_check(per) or looker == self
```
j-f1: Have you considered adding thresholds to these? For example, it's probably fairly easy to tell which race someone is, archetype may be harder, and knowing exact (rather than general) health may be even harder.

aliceafterall: I think that's an excellent idea - I actually asked about that on the issue related to this PR.

I think race should be automatic - race will be fairly easy to tell.
Archetype would be harder - since the skills that exist are relatively simple, maybe setting a higher threshold, maybe 7?

Any suggestions on the health? My thought was to make it based on medicine, not perception.

j-f1: What if you lowered the bar on the current health and required more perception/medicine points to see the character’s HP, and even more to see their total HP?

For example:

```
# 3
They seem a little roughed up.
# 5
They seem a little roughed up, and they’ve got 2 HP left.
# 7
They seem a little roughed up, and they’ve got 2/3 HP left.
```

typeclasses/characters.py
```diff
@@ -84,6 +86,101 @@ def at_pre_unpuppet(self):
...
+
+        # this is the base name format - it just colors the name cyan
+        string = "|c%s|n" % name
+
+        # if we're adding race or archetype, add "the" after the name
+        if (knows_race and self.db.archetype) or (knows_archetype and self.db.archetype):
```
j-f1: You could add these conditionals to the initialization, since if the character doesn’t have a race, I can’t know what race they are. You’re also checking `archetype` instead of `race` in the first condition.

typeclasses/characters.py
aliceafterall: That would be a good idea! Right now you can't tell what their HP is to an exact amount - perhaps a perception check for the vague descriptions, and a medicine check for exact?

typeclasses/characters.py
aliceafterall: Haha, good catch. Some unittests would have caught the mistake there, so perhaps that would be a good addition.

As for checking in the initialization, very good point. I did it this way thinking it would be more clear for someone reading it, but I think you're right.

typeclasses/characters.py
```diff
@@ -98,62 +98,92 @@ def return_appearance(self, looker):
...
+
+        knows_stamina_vague = skill_check(per, 6) or looker == self
+        knows_stamina_exact = skill_check(med, 8) or looker == self
 
         # this is the base name format - it just colors the name cyan
         string = "|c%s|n" % name
 
         # if we're adding race or archetype, add "the" after the name
-        if (knows_race and self.db.archetype) or (knows_archetype and self.db.archetype):
+        if (knows_race) or (knows_archetype):
```
j-f1: You can get rid of the parens here.

feend78:
This looks really good, @chainsol. I'd like to see a few unit tests added to cover the new perception checks.

I'm not sure what's going on with the errors in the Travis CI build. It looks as though all the tests for Ainneve are passing, but we're getting failed tests on the `evennia test evennia` part of it. I updated my local copy of the main evennia repository and I'm seeing them locally, though others in the IRC channel say they aren't seeing any. I may try deleting my local copy and checking out again, but perhaps @Griatch may also have some insight into what's going on here.

aliceafterall:
Thought so too. I'll add some basic unit tests to what's already in the return_appearance checks.

Any feedback on how "difficult" the checks should be?

feend78:
I think the target numbers you chose seem like a good starting point. The default target number for most skill checks in OA is 5, and i think having the "vague" checks a little easier and the "exact" checks somewhat harder makes a lot of sense.

feend78:
I wonder if @Griatch would be willing to check out the Travis CI build logs... It appears the errors are happening during the `evennia test evennia` part. It looks like the evennia/evennia repository has a passing Travis CI badge on it, but in my local copy of the master branch, I get the same errors as those that appear in the build log when I try to run the base framework tests. I haven't been able to figure out what is causing these errors, and the chatter in the IRC channel seems to indicate most people are not experiencing them. However, it appears that all recent Ainneve PRs (excluding #87) are failing with this same set of errors.

Griatch:
@feend78 I can confirm that the evennia unittest suite fails with the Ainneve game folder (it works with a vanilla game dir). This is most likely due to the test suite somehow making use of the normal commands and typeclasses - and those have been overridden in Ainneve's case. This was not supposed to happen since we are using `settings_defaults.py` only for the  core tests. I don't know off the bat just why there is a "leakage" between game and core, but that seems to be the case. 

I will open a new issue against Evennia concerning this. So as to not block your work here, I suggest you comment out https://github.com/evennia/ainneve/blob/master/.travis.yml#L11 so that Travis only runs Ainneve's own unit tests for now.

aliceafterall:
Just to be clear - I've written initial unittests that pass but this isn't ready to merge - need to write a little more advanced testing.

aliceafterall:
I don't know that I can improve the tests, actually - if someone wants to take a look at them that's fine, but other than testing that the exact chances are correct, I'm not sure this needs more testing - and that's a test I don't know how to write. Otherwise, I'm happy with this for the time being! Please do tell me if anything else needs changing.

aliceafterall:
I'm going to merge this into Develop - if someone wants to improve the testing, that's absolutely amazing, and I'd be excited to see another PR for that!

PR #97, opened 2017-07-21
Add a map object

j-f1:
The map will automatically fill in as you explore Ainneve. Each time you enter a room, the map will look through all of the exits representing a compass direction and store information about the rooms on the other side. The `map` command (or `look map`) will show you your map. By default, the map is also printed every time you move to a new room. The webclient has the option to show the map in a popup window, kind of like the `help` command. In addition, I’ve:

- removed an impossible exit from the meadow which really messed up the map
- added a helper function to allow specifying terrains based on the hardness levels
- added grass and road terrains
- added an `@map_tile` command to customize the tile shown on the map for a specific room
- added `Road` and `Bridge` typeclasses that render connected lines on the map. Example:
  ```
    ┌─══
  ╶─┘
  ```
  consists of:
  ```
        (road)(bridge)
  (road)(road)
  ```
- gives each character a map when they finish calling `@charcreate`

The `map` command also includes a `map/reset` command, which erases all of the data on the map, which is useful when testing the map.

Here’s the current world map:

```
┌───── MAP ────┐
│  . ~↟~ .     │
│     . .↟. .  │
│     .  .  .  │
│  ══─*─────╴  │
│     \        │
│       \^_    │
└──────────────┘
```


commands/maps.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def func(self):
+        caller = self.caller
+        maps = caller.search('map', quiet=True)
+        if len(maps) == 0:
+            caller.msg('You don’t have any maps :(')
+            return
+        theMap = maps[0]
+        if 'reset' in self.switches:
+            get_input(caller, 'Are you sure you want to wipe this map?', really_wipe_map, None, theMap)
```
aliceafterall: Perhaps document what each argument of get_input is doing? Ainneve is meant to be an example, so it might help those new to Evennia.

commands/maps.py
```diff
@@ -0,0 +1,59 @@
...
+        else:
+            caller.msg((theMap.return_appearance(self.caller), {"type": "map"}))
+
+def really_wipe_map(caller, prompt, result, theMap):
+    if result.lower() in ("y", "yes"):
+        theMap.db.map_data = {}
+        theMap.db.x = theMap.db.y = 0
```
aliceafterall: Same type of thing: It may be best to split this to two lines. I know what it does, but someone brand new to Python and Evennia might not.

commands/room_exit.py
```diff
@@ -32,8 +34,7 @@ class CmdTerrain(MuxCommand):
     Notes:
       If <room> is omitted, defaults to your current location.
 
-      Acceptable values for <terrain> are EASY, MODERATE, DIFFICULT,
-      MUD, ICE, QUICKSAND, SNOW, VEGETATION, THICKET, or DEEPWATER.
+      Acceptable values for <terrain> can be found in typeclasses/rooms.py.
```
aliceafterall: You shouldn't take out the acceptable terrain message - that's actually for the command help, we shouldn't have to read the source to know. :]

typeclasses/maps.py
```diff
@@ -0,0 +1,252 @@
...
+          a three-character string to draw on the map
+        - by specifying a `terrain` present in the TILES dictionary in maps.py
+    If no tile can be found, the map will display `?!?` as the tile.
+    '''
+    weight = 1.0
+
+    def at_object_creation(self):
+        super(Map, self).at_object_creation()
+        self.db.map_data = {}
+        self.db.x = self.db.y = 0
```
aliceafterall: Same as above - may wish to split to two lines.

typeclasses/maps.py
```diff
@@ -0,0 +1,252 @@
...
+        loc = self.location
+        if loc is None:
+            return loc
+
+        done = False
+        while 1:
+            if loc.location is not None:
+                loc = loc.location
+            else:
+                break
```
aliceafterall: I really don't like this loop - if I make a section of rooms containing themselves:
`Room1 -> Room2 -> Room3 -> Room1`
Then this will loop forever. Yes, that shouldn't happen but still.

typeclasses/maps.py
```diff
@@ -0,0 +1,252 @@
...
+            return character
+
+        done = False
+        while 1:
+            if character is None:
+                return None
+            if not character.is_typeclass('characters.Character'):
+                character = character.location
+            else:
+                break
```
aliceafterall: Same thing here, I might suggest a way to break out of the loop after some amount of iterations. I don't think this is as likely to cause problems though.

typeclasses/maps.py
```diff
@@ -0,0 +1,252 @@
...
+
+        min_x = max_x = 0
+        min_y = max_y = 0
+        for room_info in self.db.map_data.values():
+            x = room_info['x']
+            y = room_info['y']
+            if x < min_x: min_x = x
+            if y < min_y: min_y = y
+            if x > max_x: max_x = x
+            if y > max_y: max_y = y
```
aliceafterall: Lines 93 and 94 should possibly be split to two lines each.
Each of the if statements should also possibly be split:
```
if x < min_x:
    min_x = x
```
Again, for clarity and easy reading.

typeclasses/maps.py
```diff
@@ -0,0 +1,252 @@
...
+        title_line = title = u' MAP '
+        # str.center is dumb. Work around that.
+        spacer = u'─'
+        i = 0
+        while len(title_line) < width:
+            if i % 2:
+                title_line += spacer
+            else:
+                title_line = spacer + title_line
+            i += 1
```
aliceafterall: This seems like it could be simplified? It's hard to understand this way.

typeclasses/rooms.py
```diff
@@ -74,4 +92,65 @@ def mv_delay(self):
...
+      'e s': u'╔',
+      'e w': u'═',
+      's w': u'╗',
+    'e n w': u'╩',
+    'e s w': u'╦',
+    'n s w': u'╣',
+    'e n s': u'╠',
+  'e n s w': u'╬'
+}
+BRIDGE_SIDES = u'═'
```
aliceafterall, line 155: Have you tested this on various clients? I wonder how it looks on older clients.

aliceafterall (changes requested):
Just some readability and simplification things - this is awesome stuff you've written!

commands/room_exit.py
j-f1: Do you know if there’s a way to dynamically load the keys? Otherwise, the list could get out-of-date.

typeclasses/maps.py
j-f1: Hm. I could do `title.center(width, '#')` and replace all of the `#` characters with the box drawing character.

typeclasses/rooms.py
j-f1: I haven’t. This should work anywhere Unicode is supported, though. If you’d like, I could potentially add an option to use boring ASCII characters where possible.

commands/room_exit.py
aliceafterall: Good point! Let me look into it and see.

typeclasses/maps.py
aliceafterall: Hmm. That might work. Isn't there an evennia util as well?

typeclasses/rooms.py
aliceafterall: You know, rather than speculate I'll try it out and get back to you! :)

typeclasses/maps.py
j-f1: It looks like the Evennia `utils.utils.pad` just calls `str.center`.

commands/room_exit.py
j-f1: @chainsol Any updates here?

aliceafterall: Sorry, was away from my computer!
What if we used _TERRAIN.keys()?
", ".join( _TERRAIN.keys() )

j-f1:
Should I contribute this instead to the Evennia contrib? It seems like it isn’t specific to Ainneve.

aliceafterall:
I agree - it's really quite useful for most any game. If you try it on Ainneve and it works well, perhaps a contrib for Evennia?

typeclasses/rooms.py
aliceafterall: If there's a way to check if Unicode characters are able to be used, it would be good to have an alternative. If the player doesn't support Unicode characters, won't these either be stripped or display as boxes? It seems like Windows telnet doesn't support Unicode, same for some Android clients. Perhaps it doesn't matter?

j-f1: Do you know if there’s a way to tell if a session supports Unicode?

aliceafterall: Hmm. You can check the encoding setting the player sets - maybe if it's anything other than utf-8, use a fallback? I'd also suggest in the future some form of screenreader mode if you didn't already consider that.

j-f1: I’m not exactly sure what the fallback should consist of. Here’s what I’ve thought of so far:

```python
FALLBACK_CENTERS = {
        'n': '|',
        'e': '-',
        's': '|',
        'w': '-',
      'e n': '?',
      'n s': '|',
      'n w': '?',
      'e s': '?',
      'e w': '-',
      's w': '?',
    'e n w': '?',
    'e s w': '?',
    'n s w': '?',
    'e n s': '?',
  'e n s w': '+'
}
```

aliceafterall: Hmm. Would slashes ( / and \ ) work? I'm not sure about the directional crossroads though.

j-f1:
Can I get some 👀 on this @chainsol?

aliceafterall (approved):

aliceafterall:
Let me try it on my little test server, just to make sure it runs on 0.7.

aliceafterall:
Actually, I already have some concerns - I don't like the idea of modifying the webclient like this, and it already breaks the webclient.

Edit: To be clear to anyone reading this later - it's because Evennia 0.7 loads jQuery from the CDN, not the static folder.

web/template_overrides/webclient/webclient.html
```diff
@@ -0,0 +1,68 @@
...
+
+<!-- Django-template blocks available for overloading:
+    - connecting     - custom connect messages
+    - jquery_import  - for changing to a local jquery version
+    - guilib_import  - for using your own gui lib
+-->
+{% load staticfiles %}
+
+{% block jquery_import %}
+  <script src="{% static 'webclient/js/jquery-2.1.1.min.js' %}" type="text/javascript" charset="utf-8"></script>
```
aliceafterall: jQuery isn't loaded directly from the js folder - it's loaded via the code.jquery.com CDN, so this breaks the webclient

j-f1:
How does this break the webclient?

web/template_overrides/webclient/webclient.html
```diff
@@ -0,0 +1,68 @@
...
+
+<div id="mapdialog" class="dialog">
+    <div class="dialogtitle">Map<span class="dialogclose">&times;</span></div>
+    <div class="dialogcontentparent">
+        <div class="dialogcontent">
+        </div>
+    </div>
+</div>
+
+{% endblock %}
```
aliceafterall, line 65: In general I don't like the idea of adding overrides to the webclient like this - I'd want to see what @Griatch and @feend78 think of this.

aliceafterall (changes requested):
Overall everything looks good - I'm wary of the webclient having a map interface, because updates to Evennia core that improve the webclient will cause problems in the future, or we'll not get them.

I still think having a fallback for the map display is important, and you had some good thoughts on that.

aliceafterall:
It's because Evennia core, at least in 0.7, doesn't store jQuery, it's loaded from the CDN. Changing that line of course fixes it.

web/template_overrides/webclient/webclient.html
j-f1: Fixed!

j-f1:
I’m wondering if it’s possible to install the map logic by overwriting the `onText` function in a separate script. I’ll try to move the other (notification-related) change to a PR on evennia/evennia this weekend.

aliceafterall:
Nice. I'm going to wait for @Griatch or @feend78 to chime in before I approve just because I want some feedback from others about the webclient thing. I'm just wary of it...

Edit: Thinking about this, it would actually be good to be able to add "plugins" like this to the webclient anyway. Perhaps the core Evennia could support something like that...

typeclasses/maps.py
```diff
@@ -0,0 +1,237 @@
...
+        '''
+        This returns a two-dimensional list containing the map tiles.
+
+        If the owner is a builder, the map displays rows
+        containing the DB numbers for each room above each row.
+
+        If the show_player argument is `True`, the room the owning player
+        is currently in will have its middle character replaced with a `*`.
+        '''
+        player = self.current_character.player
```
aliceafterall: Ah, missed this. It should be current_character.account - since the name changed.

aliceafterall:
@j-f1 I just noticed - line 70 in typeclasses/maps.py, you check the character.player, but we'll have renamed that to character.account.

typeclasses/rooms.py
Griatch: FYI, the `UTF-8` flag is set on the client session if the client reports that it supports UTF-8. It is visible via `@options` and in code via e.g. `character.sessions.get()[0].session_flags`.

aliceafterall:
@j-f1 I'd like to merge this into Develop, is it possible to fix any references to the player to match Develop, as well as to update the dictionary we talked about, of alternate characters for non-unicode clients?

j-f1:
I’m hoping to get to it this weekend, although I might have a chance sooner than that.

aliceafterall:
Of course, no rush! I just wanted you to know I hadn't forgotten!

typeclasses/maps.py
```diff
@@ -0,0 +1,237 @@
...
+
+    @property
+    def current_character(self):
+        '''
+        This returns the character carrying the map,
+        or None if the map isn’t currently being carried by a character.
+        '''
+        return find_character(self)
+
+    def draw_map(self, show_player = True):
```
j-f1, line 60: Should I change this to something other than `show_player`? When set to `True`, it puts a `*` on the map in the room you’re in.

aliceafterall: To me, `show_player` makes sense - I don't see a problem with that naming.

web/template_overrides/webclient/webclient.html
j-f1: @Griatch or @feend78 Could you comment on this?

aliceafterall:
@j-f1 Would you say you're ready with this? The change to a separate file for the webclient changes eases my qualms with this, and I can't see anything wrong!

aliceafterall (approved):

j-f1:
@chainsol I’m good to go if you’re ready. Should I worry about [support for non-Unicode clients](https://github.com/evennia/ainneve/pull/97#discussion_r129674641)?

aliceafterall:
Yes, if you're able to, I think that would be important - sounds like Griatch suggested `character.sessions.get()[0].protocol_flags['ENCODING']` as a way to check for UTF-8/Unicode support.

aliceafterall:
@j-f1 Would it be doable to just change the map tiles to a tuple, with the unicode character and a fallback character, and then check if the session supports Unicode in the Map object itself?

j-f1:
That’s a good idea! I’ll take a look at that this afternoon. (~5-6 hours from now)

aliceafterall:
No rush! When you get to it, great, and we'll merge this in to the develop branch.

j-f1:
I’ve looked at this a few times, but I’m not sure how I should thread the result down to the `map_tile` methods.

aliceafterall:
Hmm, I've been going crazy moving, but I see your troubles there. Could the map_tile functions be passed the player trying to look at them?

Edit: Oh goodness, did you mean the `@map_tile` command? Or the map_tile method on the rooms?

j-f1:
> Oh goodness, did you mean the `@map_tile` command? Or the map_tile method on the rooms?

I meant the `map_tile` method on rooms.

aliceafterall:
So would that work? Pass down from the commands the session of the player looking at the map, and default to showing UTF-8/Unicode characters, fall-back to ASCII if the session doesn't support UTF-8?

aliceafterall:
```"
ââââ MAP ââââââ
â  . ~â~ .     â
â     . .â. .  â
â     *  .  .  â
â  ââââ¬ââââââ´  â
â     \        â
â       \^_    â
ââââââââââââââââ
```
Without having fallback characters, that's what the map looks like. I think we need to figure this out before merging it in. With a `utf-8` encoding, it looks great, though!

I took a look at it and if we changed the character dicts to be storing tuples of the "nice" character and a fallback, that might be best? That way we can check if a character is holding the map and if so, use the fallback character if the session doesn't support `utf-8`.

PR #98, opened 2017-07-24
docs: correct spelling issue

datatypevoid:


feend78:
Good catch! Thanks for the PR

PR #101, opened 2017-09-29
Upgrade to Evennia 0.7

aliceafterall:
I ran project_upgrade.py and checked the renames - everything looks good and Ainneve runs just fine - with a fresh database. If we wanted to keep the current database some manual work will have to be done there.

Caveats: Tests in world/ are failing - I'm not sure why, and this would need to be looked at, I need another pair of eyes for this one.

### DIFF
--- a/commands/chargen.py
+++ b/commands/chargen.py
@@ -1,22 +1,22 @@
 """
 Chargen commands.
 
-Ainneve implements MULTISESSION_MODE = 2, in which players enter an
+Ainneve implements MULTISESSION_MODE = 2, in which accounts enter an
 OOC state immediately after creation. From that state, the @charcreate
 command is available to create a new character. This command launches
 the Ainneve character creation EvMenu.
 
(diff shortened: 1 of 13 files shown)


aliceafterall:
Woah! This actually needs more work than I thought! It runs great - but we're using the old style color tags instead of the new style.

EDIT: It seems the custom connection screen is the only place we have the { style tags.

aliceafterall:
Our unittests are failing, however the commands work correctly. I'm working on fixing these unit tests - they'll need to be rewritten. Griatch's suggestion for testing EvMenu is to test the nodes directly - and I think that makes more sense than relying on a fragile state of a session.

Once I have these tests rewritten, this PR would be ready to merge - pending #103 either to master or a devel branch.

aliceafterall:
As this is ready to merge, but not necessarily ready for production, I'm merging it into the "develop" branch. This branch will be where the newest features and breaking changes will be made - similar to Evennia's structure.

PR #107, opened 2017-11-05
Merge Develop branch into Master

aliceafterall:
It's finally ready! The develop branch runs with no problems, all the code we needed to change is changed, and all our tests work and pass. Please read below for a summary, or [take a look at the development server.](https://ainneve.com)

## Summary
- Evennia 0.7 changed references to player to account - so most of the references here have been changed. 
- Our batchcommands used some old commands no longer in 0.7 - for instance, `@help/add/force` was changed to `@sethelp/replace` in `world/content/riverport/testing_help.ev`.
- New style color format codes instead of the old ones
- Miscellaneous testing fixes. In the future we should improve our testing suite, especially our tests of EvMenu. @Griatch added a test suite that I'll be attempting to implement here for EvMenu.

## Quirks
So far, we've tested this out fairly well. [There's a development server up](https://ainneve.com) if anyone wants to take a look. There are some bugs that Ainneve itself still has, but all the kinks related to the move to 0.7 have been fixed, as far as I've been able to find myself, and others have found.

There's an oddity with combat, where we're not clear about the fact that one needs to equip their weapon to use it, though I think that's in master as well.

## The future
- I'd like to implement the test suite for EvMenu - our current testing of EvMenu ( our chargen for instance ) is confusing, and fragile.
- I'd also like to change our tests that use hardcoded strings to actually check the result instead.
- Fix combat, or use a contrib. I'd rather the contrib path - `turnbattle` seems like it's a possibility.

### DIFF
--- a/commands/chargen.py
+++ b/commands/chargen.py
@@ -1,22 +1,22 @@
 """
 Chargen commands.
 
-Ainneve implements MULTISESSION_MODE = 2, in which players enter an
+Ainneve implements MULTISESSION_MODE = 2, in which accounts enter an
 OOC state immediately after creation. From that state, the @charcreate
 command is available to create a new character. This command launches
 the Ainneve character creation EvMenu.
 
 This new @charcreate command is added to the session cmdset. We also
 create a cmdset with the "Remove" mergetype to remove the original
-version from the player cmdset.
+version from the account cmdset.
 
-To ensure that players cannot puppet a character until it has completed
+To ensure that accounts cannot puppet a character until it has completed
 the entire character creation process, we also override the default @ic
-command on the player.
+command on the account.
 """
 from django.conf import settings
 from evennia import CmdSet
-from evennia.commands.default.muxcommand import MuxPlayerCommand
+from evennia.commands.default.muxcommand import MuxAccountCommand
 from evennia import default_cmds
 from evennia.utils import create, search
 from evennia.utils.evmenu import EvMenu
@@ -37,7 +37,7 @@ class CharCreateCmdSet(CmdSet):
 
 
 class RemoveCharCreateCmdSet(CmdSet):
-    """Command set to remove @charcreate command from Player."""
+    """Command set to remove @charcreate command from Account."""
     key = "rem_charcreate_cmdset"
     priority = 2
     mergetype = "Remove"
@@ -48,7 +48,7 @@ class RemoveCharCreateCmdSet(CmdSet):
 
 
 class ChargenICCmdSet(CmdSet):
-    """Command set to override the @ic command on Player"""
+    """Command set to override the @ic command on Account"""
     key = 'chargen_ic_cmdset'
     priority = 2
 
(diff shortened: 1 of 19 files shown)


commands/chargen.py
```diff
@@ -66,12 +66,12 @@ class CmdIC(default_cmds.CmdIC):
     Go in-character (IC) as a given Character.
 
     This will attempt to "become" a different object assuming you have
-    the right to do so. Note that it's the PLAYER character that puppets
+    the right to do so. Note that it's the ACCOUNT character that puppets
```
feend78: "ACCOUNT character" seems confusing, and is perhaps a hold-over from the original docstring. 

It could probably just be "Note that it's the ACCOUNT that puppets..."

world/test_chargen.py
```diff
@@ -15,11 +15,11 @@
 class ChargenTestCase(EvenniaTest):
     """Test case for the chargen menu process."""
     character_typeclass = Character
-    settings.PROTOTYPE_MODULES = PROTOTYPE_MODULES
 
     def setUp(self):
         super(ChargenTestCase, self).setUp()
         self.session.msg = Mock()
+        settings.PROTOTYPE_MODULES = PROTOTYPE_MODULES
```
feend78, line 22: This may not be necessary any longer, as I believe this was a workaround for using the game's settings file before the `evennia test --settings settings.py gamedir` syntax was introduced.

world/test_combat.py
```diff
@@ -494,6 +496,11 @@ def test_do_nothing(self):
 @patch('world.rulebook.utils.delay', new=CALL_IMMEDIATE)
 class AinneveCombatAttackTestCase(AinneveCombatTest):
     """Tests for attack combat actions in Ainneve."""
+    def setUp(self):
+        super(AinneveCombatAttackTestCase, self).setUp()
+        settings.PROTOTYPE_MODULES = ainneve_settings.PROTOTYPE_MODULES
```
feend78, line 501: This can likely be removed as a workaround from the era before `evennia test --settings settings.py gamedir` was supported.

feend78 (commented):
This all looks pretty good. I just made two notes about tests, where we're explicitly importing certain keys from settings.py that are probably no longer necessary now that we can specify the settings file in `evennia test`.

feend78:
One other note about the EvMenu unit tests. They are not designed to test whether the menu is rendering correctly. Rather, they are testing that the `@ic` or `@charcreate` commands called on a partially-created character **resumes the menu at the correct node**. I will agree that they have always been fragile (and therefore frustrating to work with). Just keep in mind that purpose if you plan to re-work them.

aliceafterall:
Excellent point - I forget things like this. @Griatch built a framework for testing EvMenu as well, if there is a rework with them it'd use that framework.

Edit: Just to clear up those odd `PROTOTYPE_SETTINGS` changes there, there's an odd bug in Evennia core itself, it seems. Without those, the tests will fail when they spawn items.

commands/chargen.py
aliceafterall: Yes, I think this is an artifact of using the rename utility, ha. I tried to catch everything but this slipped through!

world/test_combat.py
aliceafterall: Ah, no, actually this is needed. It is certainly possible that we shouldn't have to, but as noted below, on 0.7 we'll get a spawner error. Probably worth a bug report to Evennia core though.

feend78 (approved):
Looks good. I see the PROTOTYPE_MODULES issue has been raised in Evennia core. This is ready to go as far as I'm concerned.

aliceafterall:
Consider it done! We'll get rid of our little PROTOTYPE_MODULES hack once Evennia core fixes that.

Issue #112, opened 2018-10-08
Rework Prototypes for Evennia 0.8


aliceafterall:
Evennia 0.8 reworked the prototype system, and we make extensive use of it in our world creation code and our character generation.

To fully support Evennia 0.8, we need to rework all of our prototypes and all our code that deals with prototypes (for instance, chargen.py) to use the new prototype system. Currently prototypes are broken, though I've pushed a couple patches that should get Ainneve running on 0.8, but without being able to use the chargen menu's purchase system.

friarzen:
I think this is working well enough now to go ahead and close this issue.

PR #117, opened 2020-09-30
Fix prototypes chargen

ChrisLR:
This should address #112 .. at least partially.
It introduces new dataclasses that wrap around evennia prototypes to provide something easier to use
and fixes the chargen options to use them.
It fixes a few crashes and the item displays

### DIFF
--- a/utils/prototypeutils.py
+++ b/utils/prototypeutils.py
@@ -1,83 +1,12 @@
-from dataclasses import dataclass, field
-from evennia import spawn
-
-_NOT_FOUND = object()
-
-
-@dataclass(frozen=True)
-class PrototypeAttribute:
-    attribute_name: str
-    value: 'typing.Any'
-    category: str
-    locks: str
-
-
-@dataclass(frozen=True)
-class PrototypeDataClass:
+def create_attribute_dict(prototype):
     """
-    An utility class for wrapping Prototypes
+    Prototype dictionaries hold attributes in attrs, a list of tuples
+    We convert this to a dictionary with names as keys and attribute value as a value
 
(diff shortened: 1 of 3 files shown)


Griatch:
Opinion: Considering the intent of Ainneve is for people to read the code to learn Evennia, not sure introducing dataclasses makes for a good example of prototype usage ... it rather adds another layer of complexity to wade through when learning ...?

ChrisLR:
Its true, dataclasses might be harder to understand for beginners but I introduced them to make prototypes easier to work with in other parts of the code, particularly the attributes. The prototype attributes are given as a list of tuples which is a bit of a hassle because the code constantly accesses attributes by name. 
Would it be better if I  instead insert a new dict inside the prototype to access the attributes by name?
I could also just use  the new dict in the code that requires it, instead of using the prototype dict

friarzen:
Yes, the new dict with attributes by name seems like a much more
straight-forward approach.
The only other thing I thought about was to add a bit of guard code around
missing keys or functions that could return NoneType or similar.
Thanks for the updates, I really appreciate the switch to fstrings, and am
looking forward to merging this.

--Friarzen

ChrisLR:
Alright, I've applied the changes discussed, noticed mobs are spawned as "A normal person" but I will investigate that in another PR

friarzen (approved):

friarzen:
Ok, I've tested it and it looks good!  Thanks.  It seems to have uncovered some bugs in the later functionality (and how coins look/how the economy works).  Those don't seem to be related to any of these fixes, though.

PR #119, opened 2020-10-04
Traits test rewrite

ChrisLR:
This adds a simple function that parses lines that come from commands using Traits to display info.
Many tests are failing because they attempt to compare Strings and the order of those Strings are not Guaranteed and are prone to change. Also moved some tests to their own TestCase and/or TestMethods as tests should only have one reason to fail.


ChrisLR:
I will update more tests on another PR at a later time as I think fixing tests is the first step to get the project back on track.

Griatch (approved):
Looks fine to me, and the test still passes so that gives further credence.

PR #120, opened 2021-12-22
Add return and item tests

Jaykingamez:
Added return statement to equip.py to prevent items without action attribute to create errrors

Secondly, Added Tests for Item and Equippable typeclass


friarzen (approved):
Good catch and tests help.

PR #122, opened 2022-08-26
Replace ainneve traits by builtin contrib

ChrisLR:
As title says, to avoid keeping unnecessary custom code in ainneve.
After these changes, I was able to create a character and look at my character sheet.

**Also, I disable IRC and comment out the custom websocket setting.
It makes no sense on any setup that isnt the main server and they should be in secret_settings when on the main server.**


ChrisLR:
There are still failing tests but they seem to have been failing before this PR anyway

friarzen (approved):
Looks good, merging

PR #127, opened 2022-08-31
Overworld pt 1

ChrisLR:
This is part 1 of the overworld.

It includes the entire Overworld map taken from the Google Docs with all landmark coordinates visible.
In this PR, you need the Debug command to enter the overworld but you can move around and see the map display around you.

Next steps for the overworld would be
- Add the travel command (that uses landmarks so players dont have to spam directions)
- Implement the Landmark Exit, taking into considerations Areas may be static or dynamic (not generated yet)
- Add the prototypes for the tiles and landmarks (both rooms and exits)
- Add a way to make obvious which Landmarks have not been implemented

FYI the debug commands included should be removed or replaced in a future PR


friarzen (approved):
Everything looks good, but I have one question about typeclasses/exits.py...

Why create the extra OverworldExits at_traverse_coordinates(), instead of just overriding the default at_traverse() using kwargs and the move commands to have their usages mention the required coords?  It seems like a good opportunity to show off overriding at_traverse() ?

ChrisLR:
@friarzen at_traverse_coordinates is part of the wilderness contrib, it is already called during at_traverse. 
The problem is it did not properly check the coordinates because it relies on lockstrings for some reason.
This is a change that imho should be backported to the contrib itself, so I might eventually revert it.
Since you already approved, I will merge but we can always adjust later on.

PR #128, opened 2022-09-13
Using `xyzgrid` for towns

InspectorCaracal:
- [x] Implement Riverport as an xyzgrid map
- [x] Automatically create/update map modules in `world.maps` on server start/stop? note: implemented server creation, not start/stop
- [ ] Add a readme to `world.maps` with basic xyzgrid launcher commands and a link to the contrib docs

I have the decorated version of the map with marshes and such saved for when I figure out how to make a decorative node class and will amend it in a future PR.


ChrisLR:
Is this ready for review?

InspectorCaracal:
Mm, this rebased fine but I forgot the xyzgrid contrib has a bug right now. I'll poke at where that's coming from and see if I can get it fixed so I can wrap this up.

InspectorCaracal:
In the interest of getting this game _working_ I'm going to do the documentation stuff in a separate PR later.

PR #136, opened 2022-12-23
Added races and classes, bound them to chargen

ChrisLR:
This adds bare bones races and classes.
For now we are using dataclasses instead of Handlers but we will do that later.
We should also make better desc at some point and customize appearances but eh...


InspectorCaracal:
Looks good! Some probably-unintended opportunities at the moment due to the naive way stat choice/swapping works, but everything is sound.

InspectorCaracal (approved):

PR #138, opened 2022-12-25
Link Overworld to riverport

ChrisLR:
This allows walking from Riverport to the Overworld and back.
Going forward, each area should be properly tagged (area_key, 'area_id') and each area exit that should lead to the overworld should be tagged ('area_exit', 'area_def') so that they will be automatically linked.

It will also set a proper starting location for new characters, the fountain in Riveport.
Note that there's a contrib bug that prevents going back in Riverport that needs to be fixed before merging.

### DIFF
--- a/server/conf/at_initial_setup.py
+++ b/server/conf/at_initial_setup.py
@@ -14,8 +14,12 @@ does what you expect it to.
 
 """
 from django.conf import settings
+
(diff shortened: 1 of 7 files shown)


InspectorCaracal (approved):
This all looks good to me!

FYI, I had no problem leaving and re-entering the city; there may be an issue with your evennia installation?

PR #141, opened 2024-07-01
Some cleanup and adds levelling, basic encounter system

ChrisLR:
This PR did some cleanup and adds a first iteration of an encounter system.
The encounters are spread out into the map and respawn over time, so an area can become free of them.
Encounters near the edge of the map will be harder while those at the center will be easier.

I might change that to consider the proximity to the city instead or maybe just simplify it.

Another part of the PR is to start whittling at a levelling system and try to balance around some guidelines.

I am still working on this but I would appreciate your comments.

### DIFF
--- a/commands/combat.py
+++ b/commands/combat.py
@@ -38,7 +38,7 @@ class CmdInitiateCombat(CombatCommand):
     """Engage an opponent in combat."""
 
     key = "attack"
-    aliases = ("engage",)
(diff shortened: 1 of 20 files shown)


friarzen:
According to GitHub, this PR has merge conflicts?

ChrisLR:
@friarzen not quite sure how that happened but fixed

